PR #1, opened 2024-03-28
add matching script

jacobvjk:
closes https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/10328

- gains script `run_matching.R`, which is used to run name_match on multiple loan books in a controlled way. Parameters can be passed to underlying `r2dii.match::match_name` via `config.yml`.
- gains `example.config.yml`, which serves as a template for creating a local `config.yml` file.
- gains `expected_columns.R` which is a helper file that provides column expectations for other scripts in the repository.


example.config.yml
```diff
@@ -0,0 +1,47 @@
...
+      by_sector: TRUE
+      min_score: 0.9
+      method: "jw"
+      p: 0.1
+      overwrite: NULL
+      join_id: NULL
+    own_sector_classification:
+      use_own_sector_classification: FALSE
+      dir_own_sector_classification: "path/to/own_sector_classification_folder"
+      filename_own_sector_classification: "own_sector_classification.csv"
```
jdhoffa, line 44: NIT (and unrelated to this PR): maybe rename this to `manual_sector_classification`? 

`own` to me always sounds like `ownership`

run_matching.R
```diff
@@ -0,0 +1,194 @@
...
+}
+
+for (i in list_raw) {
+  # load raw loan books
+  raw_lbk_i <- readr::read_csv(
+    file.path(dir_raw, i),
+    col_types = col_types_raw
+  )
+
+  group_name <- gsub(".csv", "", i)
```
jdhoffa: NIt: pretty sure this could be simplified/ moved out of the loop using `vroom::vroom(..., id = "loanbook_demo")`

run_matching.R
```diff
@@ -0,0 +1,194 @@
...
+      join_id = matching_join_id
+    )
+  }
+
+  # write matched data to file
+  matched_lbk_i %>%
+    readr::write_csv(
+      file = file.path(dir_matched, glue::glue("matched_lbk_{group_name}.csv")),
+      na = ""
+    )
```
jdhoffa: NIT: In general, do you think it's preferable to write out X number of `.csv` per matched loanbook? or 1 single `matched.csv` object, with an `id` col indicating the loanbook?

jdhoffa (approved):
A couple of NIT comments, nothing blocking

run_matching.R
jacobvjk: I think it is better to have X separate `.csv` files, because I think it's easier to handle separate smaller files in the manual matching step. also makes splitting manual work between people a bit easier to handle

run_matching.R
jacobvjk: updated

example.config.yml
jacobvjk: https://github.com/RMI-PACTA/workflow.multi.loanbook/issues/4

PR #8, opened 2024-04-02
add prepare_abcd script

### LINKED ISSUE #7: feat: add script `prepare_abcd.R`
ADO: https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/10332

relates to refactoring as laid out here: https://app.mural.co/t/rmi6312/m/rmi6312/1710921788398/0538a792bff3487d254abef2dbadb7b4b041409f?sender=uc1a15eda5c426df85e617491

jacobvjk:
closes #7 

- `workflow.multi.loanbook` gains script `prepare_abcd.R`, which prepares the ABCD for ultimate use in the ensuing PACTA analysis
- gains the option `remove_inactive_companies` to remove inactive combinations of company/sector from the ABCD that have 0 production in t5 (i.e. they either stay inactive or become inactive in a given sector over the analyzed time frame, which means that calculation of aggregate metrics becomes impossible for such combinations)
- gains diagnostic file `abcd_removed_inactive_companies.csv`, which is written to `dir_abcd` when inactive companies are removed

### DIFF
--- a/prepare_abcd.R
+++ b/prepare_abcd.R
@@ -74,12 +74,14 @@ if (!all(cols_abcd$col_names_abcd %in% names(abcd))) {
 
 # optional: remove inactive companies----
 
-# remove company-sector combinations where production in t5 = 0 when
+# (1) remove company-sector combinations where production in t5 = 0 when
 # it was greater than 0 in t0.
+# (2) remove company-sector combinations where production is 0 for the entire
+# time frame from t0 to t5.
 rm_inactive_companies <- function(data,
                                   start_year,
                                   time_frame) {
-  data_no_prod_t5 <- data %>%
+  comp_sec_no_prod_t5 <- data %>%
     dplyr::filter(
       year %in% c(.env$start_year, .env$start_year + .env$time_frame)
     ) %>%
@@ -95,15 +97,13 @@ rm_inactive_companies <- function(data,
     dplyr::filter(
       !!rlang::sym(paste0("prod_", start_year)) > 0,
       !!rlang::sym(paste0("prod_", start_year + time_frame)) == 0
-    )
-
-  comp_sec_no_prod_t5 <- data_no_prod_t5 %>%
+    ) %>%
     dplyr::distinct(
       .data$name_company,
       .data$sector
     )
 
-  data_no_prod_t0_to_t5 <- data %>%
+  comp_sec_no_prod_t0_to_t5 <- data %>%
     dplyr::filter(
       year %in% c(.env$start_year, .env$start_year + .env$time_frame)
     ) %>%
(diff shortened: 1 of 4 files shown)


prepare_abcd.R
```diff
@@ -0,0 +1,165 @@
...
+# load data----
+if (!file.exists(path_abcd)) {
+  stop(glue::glue("No ABCD file found at path {path_abcd}. Please check your project setup!"))
+}
+
+abcd <- readxl::read_xlsx(
+  path = file.path(path_abcd),
+  sheet = sheet_abcd,
+  col_types = cols_abcd$col_types_abcd
+)
```
jdhoffa, line 70: NB: I know @cjyetman has been using the `tidyxl` package lately. I'm not sure why/ if there is a benefit to that over `readxl`, but just flagging in case we want to align on an approach there.

run_matching.R
```diff
@@ -113,11 +115,14 @@ if (!file.exists(path_abcd)) {
...
-  col_types = col_types_abcd,
-  col_select = dplyr::all_of(col_select_abcd)
+abcd <- readxl::read_xlsx(
+  path = file.path(path_abcd),
+  sheet = sheet_abcd,
+  col_types = cols_abcd$col_types_abcd
 )
+if (!all(cols_abcd$col_names_abcd %in% names(abcd))) {
+  stop("Columns in abcd do not match expected input names. Please check your input.")
+}
```
jdhoffa, line 125: Shouldn't you use the "prepared" abcd for matching? 

If we are  potentially removing inactive companies anyway, it could lead to a slightly easier matching process.

jdhoffa (changes requested):

run_matching.R
jdhoffa: But if not, then this LGTM :-)

jacobvjk: This is something I have been wondering for a while. On the other hand, it is not necessary to remove inactive companies from standard PACTA (it becomes a necessity for the aggregate metrics) and removing them from the start would potentially reduce the match success rate. Which may be fair, but it is not obvious to me. I think we can do that, which may indicate that both prep_* scripts in the mural should move out of the main.R script

jdhoffa: Hmm yeah that's fair. 
We can leave it as-is for now and discuss in a different forum if you like?

jdhoffa (approved):

prepare_abcd.R
cjyetman: FYI... `tidyxl` allows you to read in the data and the formats of an XLSX file, so it's a bit more capable, though a bit more complicated to use. If there's a rectangular dataset in the XLSX you're importing, `readxl` will be easier.

My other reasoning for using it is it allows me to have only the importing of a raw XLSX file in a "workflow" and leave all the transformation, cell selection, variable classing, etc. for an R function that can fully specify what is done to the raw data. This is towards the goal of keeping all "methodology" stuff in a R package, and all/any file i/o stuff in a workflow.

So that's the context for it.

jacobvjk: okay, I think in this case it is fine to keep `readxl` then for now, as the banks abcd is always rectangular. But noted that it looks pretty useful for other datasets, looks like scenarios could be a perfect use case

cjyetman: scenarios is exactly what I've been using it for 😉

PR #17, opened 2024-04-03
12 calculate match success rate

### LINKED ISSUE #12: feat: calculate match success rate

jacobvjk:
closes #12 

- `workflow.multi.loanbook` gains script run_calculate_match_success_rate.R, which calculates the match success rates in terms of...
  - matched number of loans and matched loan size
  - both in absolute and relative terms
  - for individual loan books and the meta loan book
  - and produces plots using `plot_match_success_rate()`
- gains `plots.R` which contains the relevant plotting function
- gains additional parameters documented in `example.config.yml` that allow adjusting the output format of the plots
- updates object names in `expected_columns.R` to better match their usage and updates the relevant scripts accordingly (this could have been a separate issue but given the change is very small I opted to include it here)

NOTE: `plot_match_success_rate()` should be moved to `pacta.multi.loanbook.plot` or `r2dii.plot` at some point and exported (see https://github.com/RMI-PACTA/r2dii.plot/issues/563)

### DIFF
--- a/plots.R
+++ b/plots.R
@@ -1,0 +1,121 @@
+# TODO: move to `pacta.multi.loanbook.plot` and export
+plot_match_success_rate <- function(data,
+                                    aggregate = c(TRUE, FALSE),
+                                    metric_type = c("absolute", "relative"),
+                                    match_success_type = c("n", "outstanding", "credit_limit"),
+                                    currency) {
+  # validate inputs
+  if (!inherits(data, "data.frame")) {
+    stop("Argument data must be of class data.frame. Please check your inputs.")
+  }
+  expected_cols <- c(
+    "group_id",
+    "sector",
+    "matched",
+    "match_success_type",
+    "match_success_rate",
+    "metric_type"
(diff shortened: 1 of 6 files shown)


plots.R
```diff
@@ -0,0 +1,121 @@
+# TODO: move to `pacta.multi.loanbook.plot` and export
```
jdhoffa: I sorta think this should live in `r2dii.plot`? 

Match success rate seems relevant to even a single loanbook? @MonikaFu @jacobvjk thoughts?
Would obviously be a new feature, and don't need to do it now. But in the medium term

jacobvjk: that is actually where we opened the corresponding issue, see NOTE above. But will update the comment for consistency

jdhoffa (approved):
lgtm!

MonikaFu:
Sounds like something that should live in `r2dii.plot` since it is relevant also for non-multi users, right? Happy to work on it when I get a mandate.

PR #30, opened 2024-07-30
add prepare_sector_split.R

### LINKED ISSUE #9: add prepare_sector_split.R

### LINKED ISSUE #10: apply sector split to match_prioritize

jacobvjk:
closes #9 
closes #10 

- `workflow.multi.loanbook` gains module `prepare_sector_split.R`, which calculates equal weights and primary energy based sector weights for companies
- deprecates the `"worst_case"` sector split, which was an option in the previous `workflow.aggregate.loanbooks`. This option has not been used and is unnecessarily complicated.
- beyond that deprecation, the script maintains the functionality from `workflow.aggregate.loanbooks`
- `run_match_prioritize.R` gains an option to split loan values based on outputs of `prepare_sector_split.R`. Option is set via `config.yml` 
- auxiliary data sets (`primary_energy_efficiency` and `unit_conversion`) needed to calculate the primary energy based sector split are integrated directly into `prepare_sector_split.R`. Previously, they were generated in `data-raw/`, but since we are currently not using a package structure and the data sets are not expected to be used elsewhere, they are integrated directly in the script for now.
- `prepare_sector_split.R` can remove inactive companies based on the output of `prepare_abcd.R`, in case the option `remove_inactive_companies == TRUE`


helper_functions.R
```diff
@@ -0,0 +1,69 @@
+lost_companies_sector_split <- function(abcd,
+                                        companies_sector_split) {
+  abcd_id <- abcd %>%
+    dplyr::distinct(.data$company_id, .data$name_company)
```
jdhoffa, line 4: NIT: single line pipe (i don't really care that much, just making note)

helper_functions.R
```diff
@@ -0,0 +1,69 @@
...
+
+  # identify lost_companies_sector_split and write to csv for inspection
+  lost_companies_sector_split <- companies_sector_split %>%
+    dplyr::anti_join(
+      abcd_id,
+      by = c("company_id")
+    )
```
jdhoffa, line 11: NIT: single line pipe

run_match_prioritize.R
```diff
@@ -5,14 +5,23 @@ library(readr)
 
 # source helpers----
 source("expected_columns.R")
+source("helper_functions.R")
 
 # load config----
 config_dir <- config::get("directories")
+config_files <- config::get("file_names")
+config_match_prio <- config::get("match_prioritize")
+config_prepare_sector_split <- config::get("sector_split")
```
jdhoffa, line 14: Does this add new/necessary keys to the `config`? Should that be documented somewhere?

jdhoffa (approved):
Mainly a rubber stamp/ review of syntax, but LGTM

run_match_prioritize.R
jacobvjk: no, they are already in the example.config.yml

It should be documented, but at this point I prefer getting the full workflow set up before writing an extensive documentation about these. But will open a task for it

jdhoffa: Sounds good. 

We should consider how exactly we want to slice and dice the architecture here, and what/ where things should be documented (e.g. {config.yml + scripts} could also be {function args + functions} in a package). 

The technical profile document will inform this, and ideally we can discuss it a bit more in the Tech Review.

PR #33, opened 2024-07-31
replace explicit usage of `rlang` with modern `dplyr` guidance

cjyetman:
The [modern recommendation for doing indirection](https://dplyr.tidyverse.org/articles/programming.html#indirection) with an environment variable is to use the `.data` pronoun with `[[`, and this reduces complexity a bit by not needing to use `rlang` directly.


jacobvjk (approved):
cool, I hadn't seen that. definitely a bit nicer

PR #45, opened 2024-08-12
Progress bar

cjyetman:
add a progress bar for when matching loanbooks takes a bit of time

adds dependency on {cli}, but I was already planning to suggest it to make [better error and warning messages](https://github.com/RMI-PACTA/workflow.multi.loanbook/compare/main...cli-conditions)


jacobvjk (approved):
lgtm

probably a good idea to extend this to the analysis scripts as well, once they are transferred. They can take some time as well

PR #47, opened 2024-08-13
remove `filename_raw` from example config

cjyetman:
I think `filename_raw` is no longer in use since the scripts automatically take every CSV they find in the `raw` directory?

```yml
    filename_raw: "raw_loanbook_123.csv"
```


jacobvjk (approved):
lgtm
it was mainly used to create sample loan books, but the corresponding function should probably be placed elsewhere

PR #48, opened 2024-08-13
import and adapt the `run_pacta` script

### LINKED ISSUE #19: feat: add run_pacta.R
the part that runs standard pacta in bulk for all loan books/groups

AB#11599

cjyetman:
- closes #19

### DIFF
--- a/run_pacta.R
+++ b/run_pacta.R
@@ -1,0 +1,382 @@
+# load packages----
+# library(dotenv)
+library(dplyr)
+library(pacta.multi.loanbook.analysis)
+library(r2dii.analysis)
+library(r2dii.data)
+library(r2dii.plot)
+
+
+# source helpers----
+source("expected_columns.R")
+source("plots.R")
+
+# load config----
+config <- config::get()
+
+input_path_scenario <- config$directories$dir_scenario
+input_dir_abcd <- config$directories$dir_abcd
+input_path_matched <- config$directories$dir_matched
+
+input_path_scenario_tms <- file.path(input_path_scenario, config$file_names$filename_scenario_tms)
+input_path_scenario_sda <- file.path(input_path_scenario, config$file_names$filename_scenario_sda)
+
+input_path_abcd <- file.path(input_dir_abcd, "abcd_final.csv")
+
+output_path <- config$directories$dir_output
+output_path_standard <- file.path(output_path, "standard")
+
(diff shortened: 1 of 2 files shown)


run_pacta.R
```diff
@@ -0,0 +1,382 @@
+# load packages----
+# library(dotenv)
```
jacobvjk: nit: remove

run_pacta.R
```diff
@@ -0,0 +1,382 @@
...
+
+
+# source helpers----
+source("expected_columns.R")
+source("plots.R")
+
+# load config----
+config <- config::get()
+
+input_path_scenario <- config$directories$dir_scenario
```
jacobvjk, line 16: nit: I have renamed some of the objects created from the confg in the other scripts, basically stripping the prefix `input_`. s anything that is a full path to a file is `path_* <- x/y.ext` and anything that is a directory is `dir_* <- x/y`. Not technically necessary to merge this, but maybe a good idea to use one style across the scripts.

run_pacta.R
```diff
@@ -0,0 +1,382 @@
...
+input_dir_abcd <- config$directories$dir_abcd
+input_path_matched <- config$directories$dir_matched
+
+input_path_scenario_tms <- file.path(input_path_scenario, config$file_names$filename_scenario_tms)
+input_path_scenario_sda <- file.path(input_path_scenario, config$file_names$filename_scenario_sda)
+
+input_path_abcd <- file.path(input_dir_abcd, "abcd_final.csv")
+
+output_path <- config$directories$dir_output
```
jacobvjk, line 25: nit: see above, this would be `dir_output` now

jacobvjk (approved):
a few nits related to how I ended up renaming paths and directories, but feel free to ignore if all of the config yet up gets wrapped into a dedicated function soon anyway

run_pacta.R
cjyetman: 💯 
```suggestion
```

run_pacta.R
cjyetman: 💯 agree, but I think I'll wait to do that in a separate repo-wide review/alignment

cjyetman: - see #54

run_pacta.R
cjyetman: 👍🏻 
- see #54

PR #57, opened 2024-08-14
various clean up fixes in prep for making it a package

cjyetman:
A bunch of minor fixes to clean things up before transforming this repo to a R package.
- namespace all non-base R function calls
- reduce use of `%>%` where feasible
- use `.data` and `.env` pronouns in all "data-masking" functions
- use `.data[["col_name"]]` instead of `.data$col_name` pattern which is a tiny bit safer, e.g.
    ``` r
    mtcars$mp
    #>  [1] 21.0 21.0 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 17.8 16.4 17.3 15.2 10.4
    #> [16] 10.4 14.7 32.4 30.4 33.9 21.5 15.5 15.2 13.3 19.2 27.3 26.0 30.4 15.8 19.7
    #> [31] 15.0 21.4
    mtcars[["mp"]]
    #> NULL
    ```

@jacobvjk I'm hoping this is easier to follow if I make all these minor changes at once in a separate PR, rather than interspersing all these changes with the much more dramatic changes coming with making this a package (if I don't do this before or at the same time as making it a package, R CMD build results will be total chaos).

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,13 +20,16 @@ Imports:
     dplyr,
     ggplot2,
     glue,
+    janitor,
     pacta.multi.loanbook.analysis,
     pacta.multi.loanbook.plot,
+    r2dii.analysis,
     r2dii.data,
     r2dii.match,
     r2dii.plot,
     readr (>= 2.0.0),
     readxl,
+    rlang,
     tidyr,
     withr
 Remotes:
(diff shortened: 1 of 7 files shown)


jacobvjk (approved):
lgtm

PR #58, opened 2024-08-14
improve progress bar

cjyetman:
I've noticed strange behavior with `cli::cli_progress_along()`, and apparently it is ["experimental"](https://cli.r-lib.org/reference/cli_progress_along.html).

Using the standard `cli::cli_progress_bar()` and friends seems to work as expected.


jacobvjk (approved):
lgtm

PR #66, opened 2024-08-17
test on the minimum R version supported

cjyetman:
Since we're [promising compatibility with R v4.1](https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/e6de82182b36c1c0504494bc6b2cf7a505d7d19d/DESCRIPTION#L39-L40), we should probably test that.

note: apparently the macOS w/Rv4.1 runner does not find a binary version of igraph 2.0.3, and the source version fails to compile 🤷🏻 

update: using a `macos-13` runner uses a x86_64 instead of arm64 VM, which allows the dependency installer to find binaries for R 4.1

update: pinning R to 4.1 revealed that the r2dii.data and r2dii.plot versions that were available when R 4.1 was "latest" do not contain functions used by this package, so I needed to set minimum versions for them:
- r2dii.data (>= 0.5.0) for:
  - `r2dii.data::increasing_or_decreasing`
- r2dii.plot (>= 0.4.0) for:
  - `r2dii.plot::prep_emission_intensity`
  - `r2dii.plot::prep_techmix`
  - `r2dii.plot::prep_trajectory`


jdhoffa (approved):
Thanks for the detailed explanations!! Very clear. 
LGTM

jacobvjk (approved):
lgtm

PR #68, opened 2024-08-18
replace custom stop and warn messages with `cli` versions

cjyetman:


R/run_match_prioritize.R
```diff
@@ -40,7 +45,11 @@ run_match_prioritize <- function() {
   list_matched_manual <- list.files(path = dir_matched, pattern = "^matched_lbk_.*_manual[.]csv$")
 
   if (length(list_matched_manual) == 0) {
-    stop(glue::glue("No manually matched loan book csvs found in {dir_matched}. Please check your project setup!"))
+    cli::cli_abort(c(
+      "x" = "No manually matched loan book csvs found in:",
+      " " = "{.path {dir_matched}}",
+      "i" = "Check the {.val dir_matched} parameter set in your {.file config.yml}."
```
jacobvjk: This was not meant to refer to a parameter necessarily. What I meant to suggest in the old error message was that the user should check if manually matched loan book files exist in their matched directory. Most likely reason for that not to be the case is that either the user did not do the manual validation step or named to manually validated loan books wrongly.

R/stop_if_.R
```diff
@@ -139,3 +139,59 @@ stop_if_not_expected_columns <- function(data, cols, desc = NULL) {
...
+#'
+#' @return `NULL` invisibly or an error
+#'
+#' @noRd
+
+stop_if_sector_split_not_one <- function(data) {
+  check_sector_split <-
+    dplyr::summarise(
+      data,
+      sum_share = sum(.data$sector_split, na.rm = TRUE),
```
jacobvjk:
```suggestion
      sum_share = sum(.data[["sector_split"]], na.rm = TRUE),
```

jacobvjk (changes requested):
mainly see comment on the error message

R/run_match_prioritize.R
cjyetman: good point. could you add suggested suggestions of what to check?

I'm trying to follow tidyvers guidance which is: first line is the error (what exactly was wrong), following lines are additional info (what the actual current state is versus what was expected), following lines are suggestions of what to check or do to fix it. So could be something like...

``` r
dir_matched <- "~/Desktop/test/matched"
cli::cli_abort(c(
  "x" = "No manually matched loan book csvs were found.",
  "i" = "No files matching the pattern {.code ^matched_lbk_.*_manual[.]csv$} were found in {.path {dir_matched}}. Have you done the manual matching process and named the edited CSVs properly?",
  "i" = "If {.path {dir_matched}} is not the correct directory, check the {.val dir_matched} parameter set in your {.file config.yml}."
  ))
#> Error:
#> ✖ No manually matched loan book csvs were found.
#> ℹ No files matching the pattern `^matched_lbk_.*_manual[.]csv$` were found in
#>   '~/Desktop/test/matched'. Have you done the manual matching process and named
#>   the edited CSVs properly?
#> ℹ If '~/Desktop/test/matched' is not the correct directory, check the
#>   "dir_matched" parameter set in your 'config.yml'.
```

jacobvjk: @cjyetman your suggestion here works perfectly

cjyetman: changed

jacobvjk (approved):
lgtm

Issue #54, opened 2024-08-14
review and align variable and function naming repo-wide


cjyetman:
> nit: I have renamed some of the objects created from the confg in the other scripts, basically stripping the prefix `input_`. s anything that is a full path to a file is `path_* <- x/y.ext` and anything that is a directory is `dir_* <- x/y`. Not technically necessary to merge this, but maybe a good idea to use one style across the scripts.

_Originally posted by @jacobvjk in https://github.com/RMI-PACTA/workflow.multi.loanbook/pull/48#discussion_r1716598471_

cjyetman:
@jdhoffa @jacobvjk I believe this is still relevant, but it would be a nice-to-have to ease development and understanding of the code, not a critical task.

PR #69, opened 2024-08-18
Config loading utils

cjyetman:


R/load_config.R
```diff
@@ -0,0 +1,178 @@
...
+
+get_scenario_select <- function(params) {
+  params[["project_parameters"]][["scenario_select"]]
+}
+
+get_region_select <- function(params) {
+  params[["project_parameters"]][["region_select"]]
+}
+
+get_aggregate_alignment_metric_by_group <- function(params) {
```
jacobvjk, line 66: this will have to be renamed once we tackle the more generalized grouping across the modules. but fine for now

R/prepare_sector_split.R
```diff
@@ -30,47 +30,37 @@
...
-  )
+prepare_sector_split <- function(config) {
+  config <- load_config(config)
 
-  path_advanced_company_indicators <- file.path(
-    config_prepare_sector_split$dir_advanced_company_indicators,
-    config_prepare_sector_split$filename_advanced_company_indicators
-  )
+  dir_matched <- get_matched_dir(config)
+  abcd_dir <- get_abcd_dir(config)
```
jacobvjk, line 37: nit: naming convention seems to be inversed. should be `dir_abcd`

R/run_calculate_loanbook_coverage.R
```diff
@@ -1,14 +1,13 @@
-run_calculate_loanbook_coverage <- function() {
-  # load config----
-  config_dir <- config::get("directories")
-  config_project_parameters <- config::get("project_parameters")
+run_calculate_loanbook_coverage <- function(config) {
+  config <- load_config(config)
 
-  dir_matched <- config_dir$dir_matched
+  abcd_dir <- get_abcd_dir(config)
```
jacobvjk, line 4: nit: see above

jacobvjk (approved):
lgtm
minor nits related to naming convention

R/prepare_sector_split.R
cjyetman: yeah, it's a bit mixed throughout the code... maybe I introduced that when I ported some scripts over. I was going for most-specific-first because it feels easier and faster to understand what you're looking at, especially when the names get really long. That might be a English-as-first-language bias though.

cjyetman: let's discuss naming conventions on Wednesday and then implement everywhere

cjyetman: - #54

jdhoffa: I'd like to join the naming convention call too!

PR #70, opened 2024-08-18
squelch readr msgs about columns

cjyetman:


jacobvjk (commented):
would it not be preferable to provide expected col types?

cjyetman:
> would it not be preferable to provide expected col types?

possibly, but I don't know how feasible that is because I don't know these input files very well. ❓

jacobvjk:
> > would it not be preferable to provide expected col types?
> 
> possibly, but I don't know how feasible that is because I don't know these input files very well. ❓

I added the expectations. Note that `.default = "c"` is required to allow for flexible `by_group` variables to be read in

cjyetman:
> > > would it not be preferable to provide expected col types?
> > 
> > 
> > possibly, but I don't know how feasible that is because I don't know these input files very well. ❓
> 
> I added the expectations. Note that `.default = "c"` is required to allow for flexible `by_group` variables to be read in

thanks! Eventually, we should probably get those into https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/main/R/expected_columns.R

yes, but in that case we may have to functionize them too, so they will accept variable `by_group` columns. That is really the main reason why I did not put them in there for now. Unfortunately you cannot just extend a `readr::cols_only` object like a vector, which would have made this a bit easier for now.

jacobvjk (approved):
lgtm

PR #71, opened 2024-08-18
expand paths for safety since

cjyetman:
phantomjs does not properly expand `~/` to the user's home directory, and webshot and pacta.multi.loanbook.plot don't protect against that either, so we can do it here to be safe


jacobvjk (approved):
lgtm

PR #77, opened 2024-08-20
standardize reading of raw ABCD with `read_raw_abcd()`

cjyetman:


R/read_.R
```diff
@@ -0,0 +1,24 @@
...
+        year = "numeric",
+        production = "numeric",
+        production_unit = "text",
+        emission_factor = "numeric",
+        emission_factor_unit = "text",
+        ald_timestamp = "skip"
+      )
+    )
+
+  dplyr::mutate(abcd, year = as.integer(.data$year))
```
jacobvjk, line 23: nit: were you not a proponent of adding an explicit return() at the end of functions in the past? I personally still like that, but no blocker of course

jacobvjk (approved):
lgtm
nb: one tiny nit/question

R/read_.R
cjyetman: I was! But I stopped trying to swim upstream.

PR #78, opened 2024-08-21
don't overwrite function `lost_companies_sector_split()`

cjyetman:
While it probably doesn't matter here since the function concludes immediately after, and therefore any local values get destroyed, it's probably not the safest idea to overwrite a function with the value of its result, e.g.
``` r
lost_companies_sector_split <- lost_companies_sector_split(
      abcd = abcd,
      companies_sector_split = companies_sector_split
    )

    lost_companies_sector_split %>%
      readr::write_csv(
        file = file.path(dir_matched, glue::glue("lost_companies_sector_split.csv.csv")),
        na = ""
      )
```


jacobvjk (approved):
lgtm

PR #81, opened 2024-08-21
de-duplicate the code that exports the PACTA sector outputs

cjyetman:


R/run_pacta.R
```diff
@@ -149,215 +149,45 @@ run_pacta <- function(config) {
...
+          .data[["region"]] == .env[["region_select"]],
+          .data[["sector"]] == .env[["sector_select"]]
+        ) %>%
+        nrow()
+      if (available_rows > 0) {
+        generate_individual_outputs(
+          data = results_tms_total,
+          matched_prioritized = matched_prioritized,
+          output_directory = output_path_standard,
+          target_type = "tms",
```
jacobvjk, line 182: I would expect that this does not work for all sectors, as some sectors should use `target_type = "sda"` (aviation, cement, steel)

R/run_pacta.R
```diff
@@ -149,215 +149,45 @@ run_pacta <- function(config) {
...
-        scenario = scenario_select,
-        region = region_select,
-        sector = sector_select,
-        start_year = start_year,
-        time_horizon = time_frame_select
-      )
-    } else {
-      next()
+  for (sector_select in sector_selects) {
+    for (tms_i in unique_groups_tms) {
```
jacobvjk: you probably need to call this loop twice, once for the market share sectors, once for the SDA sectors, including the corresponding variables to loop over (`sda_i in unique_groups_sda` in the latter case)

jacobvjk (changes requested):
Given the current function definition, I don't think we can get around calling the loop twice, for TMS and for SDA separately

R/run_pacta.R
cjyetman: ah crap, good catch... I'll come up with a solution

cjyetman:
@jacobvjk this might be a bit wonky (and I think make this a clear candidate for refactoring this out into a internal utility function), but should work...

https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/adbd989a36399727d821bec05693554e97603289/R/run_pacta.R#L152-L167

R/run_pacta.R
```diff
@@ -149,215 +149,48 @@ run_pacta <- function(config) {
...
+  for (vals in sector_vals) {
+    sector_select <- vals[["sector"]]
+    groups <- get(paste0("unique_groups_", vals[["method"]]))
+
+    for (i in groups) {
+      available_rows <- results_tms_total %>%
+        dplyr::filter(
+          .data[["group_id"]] == .env[["i"]],
+          .data[["scenario_source"]] == .env[["scenario_source_input"]],
+          grepl(.env[["scenario_select"]], .data$metric),
```
jacobvjk, line 172: unfortunately this still does not work, because in the tmsr results, the variable is called `metric`, whereas in sda it is called `emission_factor_metric`. So this automatically leaves 0 rows for SDA sectors

R/run_pacta.R
jacobvjk: this would still need to get `target_type = vals[["method"]]` for the SDA sectors to pass

jacobvjk (changes requested):
some persistent issues. maybe this is more complicated after all

cjyetman:
closing as this needs a serious re-think

PR #85, opened 2024-08-22
unify the 2 `remove_inactive_companies` options; keep the one in the `prepare_abcd` section

### LINKED ISSUE #84: there are two `remove_inactive_companies` options
There are two options in the config file both named `remove_inactive_companies`, though in different sections of the config so not conflicting with each other. @jacobvjk can you confirm that both are necessary, and might you have a suggestion for further distinguishing them (beyond simply existing in a different section)?

https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/bfdd711a177b3991d8397335161149a077803776/example.config.yml#L23

https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/bfdd711a177b3991d8397335161149a077803776/example.config.yml#L48-L49

cjyetman:
- closes #84 

(config loading utils already paying off 😉)

### DIFF
--- a/R/load_config.R
+++ b/R/load_config.R
@@ -48,7 +48,7 @@ get_time_frame <- function(params) {
 }
 
 get_remove_inactive_companies <- function(params) {
-  params[["project_parameters"]][["remove_inactive_companies"]]
+  params[["prepare_abcd"]][["remove_inactive_companies"]]
 }
 
 get_scenario_source <- function(params) {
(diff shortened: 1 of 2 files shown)


jacobvjk (approved):
lgtm in principle but probably good to trigger the failing check again

cjyetman:
> lgtm in principle but probably good to trigger the failing check again

yup... it was just a weird pak failure, passing now

PR #89, opened 2024-08-26
Data dictionary

### LINKED ISSUE #50: feat: gains data dictionary
AB#11667

jacobvjk:
closes #50 

- gains `data_dictionary`, which documents and explains the data structure of all outputs
- gains vignette `data_ditionary.Rmd` which provides some context for each output data set and shows the corresponding entry of the data dictionary in a `DT::datatable`

NOTE:
- column width is far from ideal in the vignette, but I wanted to get the content out first


jacobvjk:
@cjyetman @AlexAxthelm  I am a bit unsure why the R-CMD check gets hung up over no visible binding of internal data files. My understanding is that internal data should be available without further calls needed. That's how I interpret this at least: https://r-pkgs.org/data.html#sec-data-sysdata 
Also seems to work in other packages, e.g. in `pacta.multi.loanbook.analysis`. Any idea what I am missing here?

cjyetman:
> @cjyetman @AlexAxthelm I am a bit unsure why the R-CMD check gets hung up over no visible binding of internal data files. My understanding is that internal data should be available without further calls needed. That's how I interpret this at least: https://r-pkgs.org/data.html#sec-data-sysdata Also seems to work in other packages, e.g. in `pacta.multi.loanbook.analysis`. Any idea what I am missing here?

I think `R/sysdata.rda` doesn't have the objects in it yet, you just have to run `data-raw/auxiliary_data_sets.R` first

e.g.
```r
(load("~/github/rmi-pacta/workflow.multi.loanbook/R/sysdata.rda"))
#> [1] "data_dictionary"
```

jacobvjk:
> > @cjyetman @AlexAxthelm I am a bit unsure why the R-CMD check gets hung up over no visible binding of internal data files. My understanding is that internal data should be available without further calls needed. That's how I interpret this at least: https://r-pkgs.org/data.html#sec-data-sysdata Also seems to work in other packages, e.g. in `pacta.multi.loanbook.analysis`. Any idea what I am missing here?
> 
> I think `R/sysdata.rda` doesn't have the objects in it yet, you just have to run `data-raw/auxiliary_data_sets.R` first
> 
> e.g.
> 
> ```r
> (load("~/github/rmi-pacta/workflow.multi.loanbook/R/sysdata.rda"))
> #> [1] "data_dictionary"
> ```

huh correct.. this is a case of "never assume" I guess

vignettes/data_dictionary.Rmd
```diff
@@ -123,17 +123,145 @@ DT::datatable(dd_companies_included)
...
-  dd_data_timeline_bo_po
+The aggregated PACTA metrics are also run across all input banking books. The calculations produce the net aggregate alignment metric, which is defined in `pacta.multi.loanbook.analysis` and allows producing the corresponding plots using `pacta.multi.loanbook.plot`. Results are grouped at the level defined by the `by_group` parameter.
+
+### Company technology deviation
+
+For each company in the analyzed banking books, shows the deviation of the technology build-out in the final year of the analysis from the corresponding allocated scenario value. This is an intermediate result that is further processed in the calculation of the net aggregate alignment metric. Only available for sectors, which have technology level calculations using the target market share, namely `automotive, coal, oil and gas, power`.
+
+```{r dd_company_technology_deviation_tms}
+dd_company_technology_deviation_tms <- dplyr::filter(data_dictionary, .data[["dataset"]] == "company_technology_deviation_tms")
+DT::datatable(dd_company_technology_deviation_tms)
```
cjyetman, line 134: Is {DT} really necessary here? If not, probably an unnecessary complication that might add confusion. If so, possibly could use some explanation of why it's needed / what it does.

cjyetman: ack, sorry... I was thinking of {data.table}... {DT} is being used to make fancy formatted tables in the HTML output.

Should we add {DT} to Suggests for this?

cjyetman: ack! you already did 🤦🏻 never mind LOL

cjyetman (dismissed):

cjyetman (approved):
hmmm... I was skeptical because the dif file viewer said that `sysdata.rda` only changed 9 Bytes, but I just ran it locally as is and it seems to be fine

PR #91, opened 2024-08-28
fix match success rate

### LINKED ISSUE #90: bug: match_success_rate indicates 0 matches when sector split is applied
Applying the sector split updates the `id_loan` entry in `matched_prioritized` to ensure every loan remains identifiable. However, when trying to join `matched_prioritzed` with the raw loan books to determine the match success rates, this means that all `id_loan` entries are now misaligned.

jacobvjk:
closes #90 

- refactors `run_calculate_match_success_rate()` to make sub steps testable
- adds sub functions
  - `calculate_match_success_rate()`
  - `add_sectors_to_raw_lbk()`
  - `combine_raw_and_matched_loan_books()`
  - `add_match_success_rate()`
- gains test to ensure the join between raw and matched loan books needed for the calculation of the match success rate is done correctly

`combine_raw_and_matched_loan_books()` is where the bug originated. the raw and matched data sets were previously not joined correctly, because `run_match_prioritize()` changes the `id_loan`, extending it with "_<sector_abcd>_<group_id>" to facilitate identification across loan books and sector splits. `combine_raw_and_matched_loan_books()` reverses this extension locally before joining the raw and matched data sets. This allows keeping `id_loan` as a join variable


jacobvjk:
failing check seems t be due to outdated version of testthat being used in the macos-13 (4.1) check

cjyetman:
@jacobvjk I pushed a change to DESCRIPTION because `testthat::expect_in()` was added in [v3.1.9](https://testthat.r-lib.org/news/index.html?q=expect_in#testthat-319)

cjyetman (approved):

PR #103, opened 2024-09-04
formatting improvements to data_dictionary vignette

### LINKED ISSUE #101: consider formatting improvements to data dictionary vignette
a few ideas to improve the [data dictionary vignette](https://rmi-pacta.github.io/workflow.multi.loanbook/articles/data_dictionary.html):
1 - hide the `library()` and `DT::datatable()` calls in a chunk that uses `echo=FALSE`
2 - when displaying the table, maybe remove the first column which is always the same for every row, e.g. `lbk_match_success_rate`
3 - maybe point to the vignette in the [Data dictionary reference page](https://rmi-pacta.github.io/workflow.multi.loanbook/reference/data_dictionary.html)
4 - hide the row names with `DT::datatable(x, rownames = FALSE)`
5 - increase the `pagelength` so that all rows are shown on one page, instead of having to click through to a second page (all tables are < 15 rows I believe)

cjyetman:
- closes #101


jacobvjk (approved):
nit: some of the column names are still very long and lead to column widths that are too wide for the first column, taking up space that should rather be used for the `"definition"`. Is there a way to put in a hard limit for the column width of the first column and force a line break?

cjyetman:
> nit: some of the column names are still very long and lead to column widths that are too wide for the first column, taking up space that should rather be used for the `"definition"`. Is there a way to put in a hard limit for the column width of the first column and force a line break?

good idea... I'll look into it

cjyetman:
> nit: some of the column names are still very long and lead to column widths that are too wide for the first column, taking up space that should rather be used for the `"definition"`. Is there a way to put in a hard limit for the column width of the first column and force a line break?

There doesn't seem to be an easy solution to this, so I'll save it for some time in the future.

PR #105, opened 2024-09-06
use {gt} to render tables rather than {DT}

cjyetman:
Improve formatting of tables using {gt} instead of {DT}

@jacobvjk maybe try this locally with `pkgdown::build_site()` to see if you prefer this style?


jacobvjk (approved):
lgtm
this reads much more easily

PR #107, opened 2024-09-10
improve check of `by_group` parameter

jacobvjk:
see title

maybe good to wrap this in a function, given we will use it in a number of steps

### DIFF
--- a/tests/testthat/test-helper_functions.R
+++ b/tests/testthat/test-helper_functions.R
@@ -1,0 +1,14 @@
+test_that("check_and_prepare_by_group() returns expected value", {
+  expect_identical(check_and_prepare_by_group(by_group = NULL), NULL)
+  expect_identical(check_and_prepare_by_group(by_group = "NULL"), NULL)
+  expect_identical(check_and_prepare_by_group(by_group = "foo"), "foo")
+  expect_identical(check_and_prepare_by_group(by_group = "foo, bar"), c("foo", "bar"))
+})
+
+test_that("check_and_prepare_by_group() returns error when passed wrong type of input", {
+  expect_condition(check_and_prepare_by_group(by_group = 1L))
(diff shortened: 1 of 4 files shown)


jacobvjk:
not sure why the ubuntu checks fail. I'm not sure how this would be related to any of the changes.

tests/testthat/test-helper_functions.R
```diff
@@ -0,0 +1,14 @@
+test_that("check_and_prepare_by_group() returns expected value", {
+  expect_identical(check_and_prepare_by_group(by_group = NULL), NULL)
+  expect_identical(check_and_prepare_by_group(by_group = "NULL"), NULL)
+  expect_identical(check_and_prepare_by_group(by_group = "foo"), "foo")
+  expect_identical(check_and_prepare_by_group(by_group = "foo, bar"), c("foo", "bar"))
```
cjyetman, line 5: off-topic comment: I find it odd that `strsplit()` also strips out spaces (and other whitespace?) even without explicitly asking it to do so... and apparently no way to tell it not to do that

cjyetman, line 5: nit: is there any possibility of `by_group` having > length 1? I think it should still work, but could maybe test what happens with `check_and_prepare_by_group(by_group = c("foo", "bar"))`

cjyetman (approved):

tests/testthat/test-helper_functions.R
jacobvjk: there is not, unless you play around with `by_group` after reading the `config.yml`. But any string read from the `yml` using `config::get()` is returned as a vector of length one in my understanding

jacobvjk: in that sense I would prefer adding `stop_if_not_length(by_group, 1)`, if it is not `NULL`

cjyetman (approved):

PR #108, opened 2024-09-11
enable `by_group` for `run_pacta`

jacobvjk:
depends on #109 

- `run_pacta()` gains option to group the loan books by any variable specified in the `by_group` parameter
- if specified as `NULL` or `"NULL"`, will return an ungrouped result, i.e. the "meta loan book"
- number of grouping variables must be 0 or 1
- `data_dictionary` and vignettes are updated to reflect flexible `by_group` instead of fixed `group_id`


R/plots.R
```diff
@@ -146,38 +148,39 @@ generate_individual_outputs <- function(data,
...
   data <- data %>%
     dplyr::filter(
-      group_id == .env$group_id,
-      scenario_source == .env$scenario_source,
-      region == .env$region,
-      sector %in% .env$sector
+      .data[[by_group]] == .env$by_group_value,
+      .data[["scenario_source"]] == .env$scenario_source,
+      .data[["region"]] == .env$region,
+      .data[["sector"]] %in% .env$sector
```
cjyetman, line 163: nit: e.g. `.env$by_group_value` can also be transformed to `.env[[by_group_value]]`, and with `by_group` and `by_group_value` objects floating around, this might be a place where it actually makes a difference (e.g. `.env$by_group` will match `.env$by_group_value` if `by_group` is not found)

R/plots.R
```diff
@@ -146,38 +148,39 @@ generate_individual_outputs <- function(data,
...
+        dplyr::between(.data[["year"]], .env$start_year, .env$start_year + .env$time_horizon)
       ) %>%
       dplyr::mutate(
         label = dplyr::case_when(
-          .data$metric == "projected" ~ "Portfolio",
-          .data$metric == "corporate_economy" ~ "Corporate Economy",
-          .data$metric == .env$target_scenario ~ glue::glue("{r2dii.plot::to_title(toupper(.env$scenario))} Scenario")
+          .data[["metric"]] == "projected" ~ "Portfolio",
+          .data[["metric"]] == "corporate_economy" ~ "Corporate Economy",
+          .data[["metric"]] == .env$target_scenario ~ glue::glue("{r2dii.plot::to_title(toupper(.env$scenario))} Scenario")
```
cjyetman, line 183: nit: with so much of this being these `$` to `[[]]` transformations (which I support, but it's out of context), it's pretty hard to follow what this PR is actually intending to do (enabling `by_group`)

jacobvjk: I can make a separate PR just transforming $ to [[]] today, should clean this one up significantly

cjyetman:
@jacobvjk I think I got confused at some point while reviewing this, and now I don't really understand the difference between (or need for both) `by_group` and `by_group_value`. Can you explain that a bit?

jacobvjk:
> @jacobvjk I think I got confused at some point while reviewing this, and now I don't really understand the difference between (or need for both) `by_group` and `by_group_value`. Can you explain that a bit?

`by_group` is the column name of the variable you want to group the loan books by and is provided by the user. `by_group_value` is the set of values this variable can have. It is mainly used in loops over all values of `by_group` and is derived automatically.

If you set something like `by_group = "net_zero_pledge"`, then `by_group_value`s might be `"yes"` and `"no"`, indicating whether or not a bank has a net zero pledge. Some calculations will produce outputs for each of the values separately.

R/run_pacta.R
```diff
@@ -31,6 +39,10 @@ run_pacta <- function(config) {
 
   dir.create(output_path_standard, recursive = TRUE, showWarnings = FALSE)
 
+  by_group <- get_aggregate_alignment_metric_by_group(config)
+  by_group <- check_and_prepare_by_group(by_group)
+  by_group_ext <- if (is.null(by_group)) { "_meta" } else {paste0("_", by_group, collapse = "_")}
```
cjyetman: nit: I'm embarrassed to admit, I'm not sure what `paste0("_", by_group, collapse = "_")` will do without trying it. Is `collapse = "_"` basically ignored because both of the objects being pasted have length 1? If so, and that's always the case, maybe remove `collapse = "_"`?
```suggestion
  by_group_ext <- if (is.null(by_group)) { "_meta" } else { paste0("_", by_group) }
```

cjyetman (approved):

R/run_pacta.R
jacobvjk: that is correct

PR #112, opened 2024-09-12
Merge analysis pkg code

cjyetman:
This is approximately the most minimal merge of the code from the analysis package while still allowing this package to build and test without errors, warnings, or notes. Nothing is wired up to the newly merged functions here, so this package continues to rely on pacta.multi.loanbook.analysis until we're ready for that.


NAMESPACE
```diff
@@ -1,5 +1,11 @@
 # Generated by roxygen2: do not edit by hand
 
+export(aggregate_alignment_loanbook_exposure)
+export(calculate_company_aggregate_alignment_sda)
+export(calculate_company_aggregate_alignment_tms)
+export(calculate_company_tech_deviation)
+export(create_benchmark_loanbook)
```
jacobvjk, line 7: I tend to think we don't need this function

R/create_benchmark_loanbook.R
```diff
@@ -0,0 +1,217 @@
+#' Return raw loan book containing the corporate economy benchmark
```
jacobvjk, line 1: again, we can probably drop this function

data-raw/auxiliary_data_sets.R
```diff
@@ -30,8 +31,35 @@ unit_conversion <- dplyr::tribble(
   "power",       "MWh",             8.598e-08
 )
 
+
+# activity units
+
+library(dplyr)
```
jacobvjk, line 37: I wonder if this `library()` call was in there for the pipes.. maybe a good reason to remove the pipes and get rid of this call? seems a bit awkward

cjyetman: yup, there just for the pipes... and I left it there rather than make more dramatic changes to the existing code I was copying in

NAMESPACE
cjyetman: remove the function completely or just not export it?

jacobvjk: It's not being used anywhere, so I would opt for removing it entirely

data-raw/auxiliary_data_sets.R
jacobvjk: okay, maybe a follow up issue

jacobvjk (approved):

PR #122, opened 2024-09-16
merge plotting code from pacta.multi.loanbook.plot

cjyetman:
⚠️ This will fail CI on one Note about depending on more than 20 packages, but this will disappear once the new internal functions are wired up and there's no longer a dependency on pacta.multi.loanbook.plot and pacta.multi.loanbook.analysis

Note: I did *not* port over pacta.multi.loanbook.plot's `deparse_1()` which is a [backport of `base::deparse1()` for R < 4.0.0](https://github.com/RMI-PACTA/pacta.multi.loanbook.plot/blob/dd14a145ac61b3b0b6446612742b51a9926c9125/R/utils.R#L7C2-L11) because this package already [depends on R >= 4.1.0](https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/67d8085432b3e171235475050691935c14dbf600/DESCRIPTION#L53) (and I converted the handful of uses of `deparse_1()` to `deparse1()`)


jdhoffa (approved):
From what I can tell, this LGTM!

R/plot_scatter.R
```diff
@@ -0,0 +1,235 @@
...
+  if (data_level == "company") {
+    p <- p +
+      ggplot2::guides(
+        shape = "none"
+      )
+  }
+
+  p
+}
+# nolint end
```
jacobvjk: This makes me wonder if we are bringing `lintr` over as well

jacobvjk (approved):
lgtm

nit: if we are not using lintr here yet we may want to remove the `#nolint` comments

R/plot_scatter.R
cjyetman: let's not 😆

cjyetman:
also should add Monika to DESCRIPTION while absorbing all this code... will do that now

PR #125, opened 2024-09-17
wire in new internal functions and drop dependency on former packages

### LINKED ISSUE #124: merge `validate_data_has_expected_cols()` and `stop_if_not_expected_columns()`
`validate_data_has_expected_cols()` was brought over from pacta.multi.loanbook.analysis while there was already `stop_if_not_expected_columns()` here. We should merge their functionality and convert any calls to `validate_data_has_expected_cols()` in the ported code from pacta.multi.loanbook.analysis to `stop_if_not_expected_columns()`.

cjyetman:
- closes #124

completes the process of fully merging the formerly external analysis and plotting pkgs


cjyetman:
@jacobvjk not sure how to modify this line in the docs. Any idea?

https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/ab0bfc003b9c27c54a98bb3f28974ac4a1bc9201/vignettes/data_dictionary.Rmd#L179

jdhoffa (approved):
looks great to me!

jacobvjk:
> @jacobvjk not sure how to modify this line in the docs. Any idea?
> 
> https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/ab0bfc003b9c27c54a98bb3f28974ac4a1bc9201/vignettes/data_dictionary.Rmd#L179

@cjyetman I would do this.

> ... which is defined in the vignettes "Calculation of a company alignment metric" and "Calculation of exposure-weighted aggregated alignment metric" and allows producing the corresponding plots. Results are grouped at the level defined by the `by_group` parameter.

jacobvjk (commented):
lgtm in principle, but I am not convinced we should not export the individual functions for the net alignment metric and the plots. If we do this, we prevent the users from potentially writing their own workflow and/or using combinations that we may not allow setting in this package.

I guess there are two ways to think about it:
- Either we say that they should not be able to access the individual functions because we give them a one stop shop experience and everything should be handled via the config. In that case we need to decide if we want to surface more options in the params and or create even more results by default (i.e. all reasonable combinations)
- Or we say that there may be power users that have very specific needs that we do not want to cover in the main workflow-y functions so keep the function newly added functions exported for the advanced users to tinker with.

An additional thought on that: if we decide not to export them anymore - which is a perfectly fair outcome - I believe we should reconsider which files produce as outputs and if we document data sets in the data dictionary that would have to be considered "intermediate".

Any thoughts on this @cjyetman  @jdhoffa ?

jdhoffa:
> Any thoughts on this @cjyetman @jdhoffa ?

My thoughts on this are: 
* To start with, export the minimal number of useful functions
* Get everything working, tested, documented, and spick and span
* Get it on CRAN
* Explore exporting additional functionality

Exploring more bespoke use-cases, and custom functionality (e.g. by exposing these lower-level functions) seems like a great idea for a second release, once we have a proven user-base for the "default/base" functionality. But at this point, more exports just means higher maintenance burden, which is not worth at this stage in development IMO

jacobvjk:
> > Any thoughts on this @cjyetman @jdhoffa ?
> 
> My thoughts on this are:
> 
> * To start with, export the minimal number of useful functions
> * Get everything working, tested, documented, and spick and span
> * Get it on CRAN
> * Explore exporting additional functionality
> 
> Exploring more bespoke use-cases, and custom functionality (e.g. by exposing these lower-level functions) seems like a great idea for a second release, once we have a proven user-base for the "default" functionality.

Just to be clear, the lower level functions are already documented and tested. But that works for me.

cjyetman:
> lgtm in principle, but I am not convinced we should not export the individual functions for the net alignment metric and the plots. If we do this, we prevent the users from potentially writing their own workflow and/or using combinations that we may not allow setting in this package.
> 
> I guess there are two ways to think about it:
> 
> * Either we say that they should not be able to access the individual functions because we give them a one stop shop experience and everything should be handled via the config. In that case we need to decide if we want to surface more options in the params and or create even more results by default (i.e. all reasonable combinations)
> * Or we say that there may be power users that have very specific needs that we do not want to cover in the main workflow-y functions so keep the function newly added functions exported for the advanced users to tinker with.
> 
> An additional thought on that: if we decide not to export them anymore - which is a perfectly fair outcome - I believe we should reconsider which files produce as outputs and if we document data sets in the data dictionary that would have to be considered "intermediate".
> 
> Any thoughts on this @cjyetman @jdhoffa ?

So, I removed the exports because, as I understood it, once they became internal to this package there was no longer any purpose for them to be exported, i.e. they were only exported before because they were contained in a different package so had to be exported to be useful here in this package.

I'm 100% indifferent on whether they are exported or not (other than in general I'm aligned with @jdhoffa on only exporting when you're sure it's needed because it's harder to "un-export" something). I do not know any use case for them myself, but if you think they are useful, I don't really have an opinion on it.

jdhoffa:
Totally, and it makes sense to keep all those docs and tests, I just think we shouldn't necessarily export them, at least for now :-) 

And something to re-consider a little closer to release I think.

cjyetman:
> > @jacobvjk not sure how to modify this line in the docs. Any idea?
> > https://github.com/RMI-PACTA/workflow.multi.loanbook/blob/ab0bfc003b9c27c54a98bb3f28974ac4a1bc9201/vignettes/data_dictionary.Rmd#L179
> 
> @cjyetman I would do this.
> 
> > ... which is defined in the vignettes "Calculation of a company alignment metric" and "Calculation of exposure-weighted aggregated alignment metric" and allows producing the corresponding plots. Results are grouped at the level defined by the `by_group` parameter.

done... with linkage to those vignettes

vignettes/data_dictionary.Rmd
```diff
@@ -176,7 +176,7 @@ plot_table(table)
 
 ## Aggregated PACTA metrics
 
-The aggregated PACTA metrics are also run across all input banking books. The calculations produce the net aggregate alignment metric, which is defined in `pacta.multi.loanbook.analysis` and allows producing the corresponding plots using `pacta.multi.loanbook.plot`. Results are grouped at the level defined by the `by_group` parameter.
+The aggregated PACTA metrics are also run across all input banking books. The calculations produce the net aggregate alignment metric, which is defined in the vignettes ["Calculation of a company alignment metric"](company_alignment_metric.html) and ["Calculation of exposure-weighted aggregated alignment metric"](loanbook_aggregated_alignment_metric.html) and allows producing the corresponding plots. Results are grouped at the level defined by the `by_group` parameter.
```
jacobvjk, line 179: will this `loanbook_aggregated_alignment_metric.html` link to anything? within the package, we do not have the html files, so I am assuming that gitpages understands that this is the fully rendered vignette, even though we do not provide the full url?

jacobvjk (approved):
lgtm

vignettes/data_dictionary.Rmd
cjyetman: it works in the pkgdown site (I tested locally) because all the vignettes are rendered to HTML in the same directory, so a "relative path/URL" is sufficient

I could not figure out any other way to link to a vignette other than using e.g. `vignette("data_ditionary")` as I did elsewhere in this PR, which renders in the pkgdown site exactly like that, and I thought the name of the vignette properly spelled out was important here.

Would very much agree to a better/alternate way of doing this, but I couldn't find one in my searching today. 🤷🏻

PR #127, opened 2024-09-17
Enable by group lbk coverage

jacobvjk:
- gains option to produce grouped loanbook coverage calculations based on `by_group` parameter

Note:
I kept this PR quite minimal. The function would benefit from some refactoring to facilitate testing of the relevant underlying calculations, but in the spirit of keeping this reviewable I decided to do that in follow up tasks instead. (#129 , #128 )


cjyetman:
(depends on #125 to pass CI without Notes that cause failures)

R/run_calculate_loanbook_coverage.R
```diff
@@ -40,12 +44,20 @@ run_calculate_loanbook_coverage <- function(config) {
...
+  if (is.null(by_group)) {
+    by_group <- "meta"
+    matched_prioritized <- matched_prioritized %>%
+      dplyr::mutate(meta = "meta")
+  }
+
   matched_companies <- matched_prioritized %>%
     dplyr::distinct(
-      .data$group_id,
+      !!rlang::sym(by_group),
```
cjyetman: nit: is this less complicated?
```suggestion
      .data[[by_group]],
```

cjyetman: not 100% sure my suggestion works

cjyetman (approved):

R/run_calculate_loanbook_coverage.R
jacobvjk: it does, I thought I had tried this, but I must have been doing something else

PR #138, opened 2024-09-18
Improve data masking

### LINKED ISSUE #128: data masking: use `.data[["var"]]` instead of `.data$var`(same for `.env`)
more robust

jacobvjk:
closes #128


jdhoffa:
looks robust to me

jacobvjk:
actually just saw there ae some more cases in the tests... will tackle those too, just so it's done

cjyetman (approved):
wow, nice! ❤️

jacobvjk:
@cjyetman I updated the test files too

PR #141, opened 2024-09-18
combine man pages of `plot_*` and `prep_*` function pairs

### LINKED ISSUE #135: consider grouping man pages for matching `plot_*()` and `prep_*()` functions
consider grouping man pages for matching `plot_*()` and `prep_*()` functions

Every `plot_*()` function is paired with a `prep_*()` function. It would make sense to combine the documentation of each pair into a single man page to reduce clutter.

technique explained here:
https://roxygen2.r-lib.org/articles/reuse.html#multiple-functions-in-the-same-topic

related to (but different than) #119

cjyetman:
- resolves #135


man/plot_sankey.Rd
```diff
@@ -12,6 +13,15 @@ plot_sankey(
...
 )
+
+prep_sankey(
+  data_alignment,
+  region,
+  year,
+  group_var,
+  middle_node,
+  middle_node2 = NULL
+)
```
jdhoffa, line 24: Style: It seems a bit counter-intuitive that `prep_*` comes after `plot_*` in the documentation here. What do you think?

cjyetman: I did that because in my mind the `plot_*` function is the "real" function, or the most important function, or at least the primary function that we most want to talk about, and the `prep_*` function is just a helper for it?

jdhoffa: If it's intentional then makes sense! just wanted to make sure it wasn't accidental.

Approved, although curious what @jacobvjk thinks too?

jdhoffa (approved):

man/plot_sankey.Rd
jacobvjk: tbh I think it doesn't really matter which comes first. If we do not export these functions, there is no "primary" function for me. In a sense, if we decide at some point that we do not want to export these functions at all, we may even scrap the manual for the prep_* function, no?
I think for now it makes sense to keep both, because we may want to export them again at some point and they are basically ready for that. long story short: fine for me

jacobvjk (approved):

PR #153, opened 2024-09-23
use prepared abcd in matching

### LINKED ISSUE #14: Shouldn't you use the "prepared" abcd for matching?
Shouldn't you use the "prepared" abcd for matching? 

If we are  potentially removing inactive companies anyway, it could lead to a slightly easier matching process.

_Originally posted by @jdhoffa in https://github.com/RMI-PACTA/workflow.multi.loanbook/pull/8#discussion_r1549115552_
            

AB#10475

Advantage would be to slightly improve matching performance and use one consistent ABCD data set throughout the analysis.

Drawback is that the removal of inactive companies is only needed for aggregate metrics, not standard PACTA and applying the prep/removal will add an unnecessary step for standard PACTA potentially reducing the match success rate.

jacobvjk:
closes #14 

- `run_matching()` and `run_match_prioritize()` now both use the `abcd_final.csv` as input for the abcd, instead of the raw abcd file
- this means that `abcd_final` now simply has to be prepared once using `prepare_abcd()` as the very first step of the analysis and all following functions expect that same data set
- this is a prerequisite for implementing the information architecture as agreed upon in https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/11852

### DIFF
--- a/R/run_match_prioritize.R
+++ b/R/run_match_prioritize.R
@@ -2,8 +2,7 @@ run_match_prioritize <- function(config) {
   config <- load_config(config)
 
   dir_matched <- get_matched_dir(config)
-  path_abcd <- get_abcd_path(config)
-  sheet_abcd <- get_abcd_sheet(config)
+  abcd_dir <- get_abcd_dir(config)
 
   match_prio_priority <- get_match_priority(config)
 
@@ -11,6 +10,10 @@ run_match_prioritize <- function(config) {
   sector_split_type_select <- get_sector_split_type(config)
 
   # validate config values----
+  stop_if_not_length(abcd_dir, 1L)
+  stop_if_not_inherits(abcd_dir, "character")
+  stop_if_dir_not_found(abcd_dir, desc = "ABCD data")
+
   stop_if_not_length(dir_matched, 1L)
   stop_if_not_inherits(dir_matched, "character")
   stop_if_dir_not_found(dir_matched, desc = "Matched loanbook")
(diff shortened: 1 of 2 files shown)


R/run_match_prioritize.R
```diff
@@ -2,15 +2,18 @@ run_match_prioritize <- function(config) {
...
   match_prio_priority <- get_match_priority(config)
 
   apply_sector_split <- get_apply_sector_split(config)
   sector_split_type_select <- get_sector_split_type(config)
 
   # validate config values----
+  stop_if_not_length(abcd_dir, 1L)
+  stop_if_not_inherits(abcd_dir, "character")
+  stop_if_dir_not_found(abcd_dir, desc = "ABCD data")
+
```
cjyetman: maybe you want to check that the file exists in the specified directory (also below)?
```suggestion
  stop_if_dir_not_found(abcd_dir, desc = "ABCD data")
  stop_if_file_not_found(file.path(abcd_dir, "abcd_final.csv"), desc = "ABCD final")
```

R/run_matching.R
```diff
@@ -23,9 +22,9 @@ run_matching <- function(config) {
   stop_if_not_inherits(dir_raw, "character")
   stop_if_dir_not_found(dir_raw, desc = "Raw loanbook")
 
-  stop_if_not_length(path_abcd, 1L)
-  stop_if_not_inherits(path_abcd, "character")
-  stop_if_file_not_found(path_abcd, desc = "ABCD data")
+  stop_if_not_length(abcd_dir, 1L)
+  stop_if_not_inherits(abcd_dir, "character")
+  stop_if_dir_not_found(abcd_dir, desc = "ABCD data")
```
cjyetman, line 27: and here?
```suggestion
  stop_if_dir_not_found(abcd_dir, desc = "ABCD data")
  stop_if_file_not_found(file.path(abcd_dir, "abcd_final.csv"), desc = "ABCD final")
```

cjyetman (approved):

PR #156, opened 2024-09-24
Update I/O paths to new information architecture

### LINKED ISSUE #22: overhaul I/O directory structure based on information architecture

jacobvjk:
closes #22 

- all functions expect only two main directories from `config.yml`: `dir_input` and `dir_output`
- updates `example.config.yml` to reflect new I/O config requirements
- updates `vignettes/config_yml.Rmd` to reflect the new I/O structure
- updates `R/load_config.R` so that the functions return the directories and paths as required under the new IA
- updates references in error messages to point to the correct paths under the new IA
- `dir_input` contains all files that the user must prepare prior to the analysis: raw loan books (in the subdirectory `"input/loanbooks"`), ABCD, scenarios, potentially company ids for the sector split, potentially a manually prepared sector classification system, which looks as follows (sub directory must be created by the user)
```
- input
  - loanbooks
```

- `dir_output` contains all files created in the process of running the P4S analysis and creates a sub-directory for each exported function, so that the user will find the outputs of the process they just ran in the corresponding sub-directory. This looks as follows (sub directories are automatically created as side effects by calling the exported functions)
```
- output
  - prepare_abcd
  - matched_loanbooks
  - prioritized_loanbooks_and_diagnostics
  - analysis
```


R/run_match_prioritize.R
```diff
@@ -41,32 +44,32 @@ run_match_prioritize <- function(config) {
...
-  list_matched_manual <- list.files(path = dir_matched, pattern = "^matched_lbk_.*_manual[.]csv$")
+  list_matched_manual <- list.files(path = output_matched_loanbooks_dir, pattern = "^matched_lbk_.*_manual[.]csv$")
 
   if (length(list_matched_manual) == 0) {
     cli::cli_abort(c(
       "x" = "No manually matched loan book csvs were found.",
-      "i" = "No files matching the pattern {.code ^matched_lbk_.*_manual[.]csv$} were found in {.path {dir_matched}}. Have you done the manual matching process and named the edited CSVs properly?",
-      "i" = "If {.path {dir_matched}} is not the correct directory, check the {.val dir_matched} parameter set in your {.file config.yml}."
+      "i" = "No files matching the pattern {.code ^matched_lbk_.*_manual[.]csv$} were found in {.path {output_matched_loanbooks_dir}}. Have you done the manual matching process and named the edited CSVs properly?",
+      "i" = "If {.path {output_matched_loanbooks_dir}} is not the correct directory, check the {.val dir_input} parameter set in your {.file config.yml}."
```
cjyetman: seems like the directory that has matched loanbooks is in the output directory, not the input directory

jacobvjk: that is correct and expected. They are the output of one of the steps covered by these functions, which falls into our definition of outputs more broadly

cjyetman:
```suggestion
      "i" = "If {.path {output_matched_loanbooks_dir}} is not the correct directory, check the {.val dir_output} parameter set in your {.file config.yml}."
```
than this error message should reference the output dir instead of the input dir

jacobvjk: wouldn't it make sense to do this instead?
```
stop_if_no_files_found(list_matched_manual, output_matched_loanbooks_dir, "dir_output", "manually matched loan book CSVs matching the pattern {.code ^matched_lbk_.*_manual[.]csv$}")
```

cjyetman: even better 😃

cjyetman (approved):

Issue #104, opened 2024-09-04
Expand cookbook


jacobvjk:
We currently have some documentation that will be part of the cookbook

- simple example of running the code
- config setup
- data dictionary

The final cookbook should be usable as a central document based on which the user can implement the entire analysis and/or find help in other resources.

The suggested structure is:

- [x] Title page/Intro/Context
- [x] Preparatory steps
  - [x] Which data sets are required as input?
    - [x] Where/how to obtain external inputs
    - [x] How to prepare own inputs (loan books and possibly sector classification)
    - [x] Highlight which ones are required and which ones are optional
  - [x] Installing relevant software
    - [x] R, RStudio, this package, dependencies
  - [x] Setting up the project
    - [x] configuration file
    - [x] Folder structure
- [x] Running the analysis
  - [x] Structure of the flow of the analysis
  - [x] Data prep
    - [x] ABCD preparation
    - [x] Optional: sector split
  - [x] Matching
    - [x] Matching the loan books
    - [x] Manual validation step
  - [x] Prioritize and Diagnose
    - [x] Prioritizing the manually matched loans
    - [x] Checking coverage and iterating in a smart way
    - [x] misclassified loans
  - [x] Run main analysis
- [x] Interpretation of outputs
  - [x] Types of results: coverage, standard PACTA, aggregate alignment metric
  - [x] Sections for each
    - [x] Example plots
    - [x] Interpretation
    - [x] Data dictionary
    - [x] Mapping of dd to graph
- [x] Advanced use cases
  - [x] How to adjust the config and inputs to generate the results of interest
  - [x] gather realistic use cases for the analysis and add one page each

jacobvjk:
@cjyetman @jdhoffa I think in the end a cookbook that can be used as the main information hub for users to run P4S by themselves could have a structure like this. Feel free to comment any additions or aspects that you think do not belong here. I realize these are a lot of bullets, but in some cases multiple of those be may combined into a single page.

jdhoffa:
I think the structure looks great! I have some comments (might be things you are already thinking of, but good to have them explicit!). Also not necessarily saying each of these comments is strictly necessary, just brainstorming 😄 

I'll start with General suggestions, and then more specific suggestions per section.

# General suggestions
**Modularity**
Consider making each section as modular as possible, so users can jump to the section they need without following the entire document in order.

**Troubleshooting and FAQs**
[...]

PR #167, opened 2024-10-14
Cookbook run analysis

jacobvjk:
relates to: https://github.com/RMI-PACTA/pacta.multi.loanbook/issues/104
the check points of the section "Running the Analysis" are addressed in this PR

- gains cookbook section on running the analysis
  - this contains a brief overview of the required functions for running the analysis
  - it briefly recaps the setup before running the functions (loading the package, setting the config path)
- the cookbook gains descriptions of...
  - how to run each function
  - how to set options for each function
  - the required intermediate step of manually validating the matches
  - options available for handling misclassified loans
  - elaborates on how often a user can expect to run each function and why
- the cookbook gains links to additional material that helps the user understand details of the process that are better explained elsewhere
  - details on matching, especially manual matching, which can be found on the PACTA for Banks website
  - References of functions that are called from r2dii.match and for which the config simply passes arguments to the underlying functions as is


cjyetman:
Are the changes to `vignettes/config_yml.Rmd` intentional, and if so, can you briefly explain why they are needed because it's very hard to tell what the impact will be from the markdown diff alone, but I know getting the levels right over such a large document can be difficult?

jacobvjk:
> Are the changes to `vignettes/config_yml.Rmd` intentional, and if so, can you briefly explain why they are needed because it's very hard to tell what the impact will be from the markdown diff alone, but I know getting the levels right over such a large document can be difficult?

Ah yea, forgot to mention. While referencing some of the points there, I realized that some of the levels were off. But maybe better to fix in another PR. I will remove these changes here

cjyetman (approved):

PR #172, opened 2024-10-15
Interpretation of results

jacobvjk:
relates to: #104 

- gains section "Interpretation of results"
- for each relevant output, this section
  - gives a brief overview
  - provides example plots if plots exist for the output
  - explains how to interpret the plot
  - shows the underlying data structure using the data dictionary
  - explains how the values from the data dictionary map to the plot


jacobvjk:
seems like the size of the installed package does not go down well with GH.. might try to use lower resolution versions of the screenshots for the cookbook. ready to review otherwise

jacobvjk:
> seems like the size of the installed package does not go down well with GH.. might try to use lower resolution versions of the screenshots for the cookbook. ready to review otherwise

compressing the png files resolved the issue

cjyetman (approved):

PR #177, opened 2024-10-21
only run net aggregate alignment when intersection of lbk and scenario sectors greater 0

### LINKED ISSUE #176: test that at least one matched asset for the given methodology is available so that it doesn't error
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/96efb02beb3fb6ca75b94d6118a431d0d0eacb7c/R/calculate_company_alignment_metric.R#L400

these lines would have to do something appropriate if `calculate_company_aggregate_alignment_sda()` doesn't find any appropriate matches/returns an empty data frame or whatever
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/96efb02beb3fb6ca75b94d6118a431d0d0eacb7c/R/run_aggregate_alignment_metric.R#L177-L183

AB#12038

jacobvjk:
closes #176 

- `run_aggregate_alignment_metric()` of the `analyse()` step gains logic to only return result for either tms or sda related net aggregate alignment calclations, if there is an overlap of matched sectors in the loan books and available sectors in the relevant scenario. If no such overlap exists, the calculation is skipped and an informative message is printed.
- E.g.: if we can only match aviation loans in the SDA related sectors, but the selected scenario contains only pathways for cement and steel, the SDA related results will be skipped with a message similar to: `"None of the sectors available for SDA analysis in the nze_2050 scenario of weo_2022 are present in the matched and prioritized loan book. Calculation of aggregated SDA results not possible. Skipping!"`
- Given that #179 ensures that any files that may exist from past runs of the analysis are completely removed if the target directory remains the same, there is no chance to accidentally retain results from different runs anymore and therefore no need to write empty result tables in case a calculation cannot be run (which was suggested in a previous version of this PR)


jacobvjk:
@cjyetman not exactly the most elegant solution and I am unsure if it is preferable to return empty results where the combinations are missing or none. But this should resolve the issue we saw in the integration test

R/run_aggregate_alignment_metric.R
```diff
@@ -129,31 +147,36 @@ run_aggregate_alignment_metric <- function(config) {
...
   company_alignment_bo_po_tms <- company_technology_deviation_tms %>%
     calculate_company_aggregate_alignment_tms(
       scenario_source = scenario_source_input,
       scenario = scenario_select,
       level = "bo_po"
     )
+  } else {
+    company_technology_deviation_tms <- tibble::tibble()
+    company_alignment_net_tms <- tibble::tibble()
+    company_alignment_bo_po_tms <- tibble::tibble()
```
cjyetman: I think a completely empty tibble might be the wrong thing here. I'd rather see an "empty" tibble, i.e. 0 rows, but that has the same structure as the intended tibble, i.e. the same column names and possibly types. If not that, than it might as well be `NULL` I think, or?

jacobvjk: out of those options, I would tend to prefer empty tibbles with headers I think

jdhoffa:
As I understand it, the purpose of this PR is to add some input expectations to various function inputs and provide sensible fall-backs in case there is no sensible input data. The sensible fallback chosen is to write empty (0-row) tibbles that contain the expected output column structure of the function. 

How will downstream code handle and/or make use of these 0-row tibbles? I see that they are written to disk in the output directory. Will subsequent plotting functions handle and/or error gracefully if input with a 0-row tibble?

I can see two options here. After a function receives input data that it cannot work with it could: 
1 - write an empty tibble to disk -> every subsequent function needs to then consider how to operate with an empty tibble input
2 - write nothing to disk -> every subsequent function needs to check if it's crucial inputs exist in the expected directory prior to running

To me, the second option seems slightly cleaner and less confusing (if a user wishes to inspect their output directory, they won't have to make sense of "empty" data frames written to disk)

 I'm curious if there was a reason not to take this approach?

cjyetman (approved):

jdhoffa:
I believe given that #179 has been merged, that we now may want to reconsider the behaviour we introduce in this PR? 
Specifically, I would like to revisit: 

> The decision to return empty files rather than not returning anything is made because running the analysis() function with the same set of parameters twice will try to overwrite past results. If, for some reason, a valid result file was created in a past run that can now not be recreated because of this lack of overlap, not writing anything would retain the old result. This could easily lead to confusion and falsely connecting results from two different runs

My preference here would be to NOT write files if those files don't have any data, rather than writing a bunch of empty tibbles. 

What are others thoughts here?

jacobvjk:
> I believe given that #179 has been merged, that we now may want to reconsider the behaviour we introduce in this PR? Specifically, I would like to revisit:
> 
> > The decision to return empty files rather than not returning anything is made because running the analysis() function with the same set of parameters twice will try to overwrite past results. If, for some reason, a valid result file was created in a past run that can now not be recreated because of this lack of overlap, not writing anything would retain the old result. This could easily lead to confusion and falsely connecting results from two different runs
> 
> My preference here would be to NOT write files if those files don't have any data, rather than writing a bunch of empty tibbles.
> 
> What are others thoughts here?

already working on simplifying it, i.e. not writing anything in those cases and reurning a message instead. will update the description of the PR in a second

jdhoffa:
Thank you!!

cjyetman (approved):

PR #180, opened 2024-10-30
check that required input files for plots are avaiable or skip

jacobvjk:
plotting of the net aggregate alignment metric requires results from the previous step to exist and not be NULL
These checks ensure the required results are available or skip the creation of the relevant plot

- returns `NULL` in `run_aggregate_alignment_metric()`, when no tms or sda result can be calculated, to allow the rest of the process to continue
- adds checks for existence of required input files for `plot_aggregate_alignment()`
- updates checks in `plot_aggregate_alignment()` to ensure inputs are not `NULL` and have some usable content
- wherever an input is missing for a result that does not hinder the remainder of the function to produce useful outputs, the relevant section is skipped
- some old checks in `plot_aggregate_alignment()` are simplified and some obsolete conditions are removed

### DIFF
--- a/R/plot_aggregate_loanbooks.R
+++ b/R/plot_aggregate_loanbooks.R
@@ -139,7 +139,7 @@ plot_aggregate_loanbooks <- function(config) {
   ### sankey plot----
   # TODO: when benchmarks get re-introduced, they need to be removed here
   # Plot sankey plot of financial flows scenario alignment - examples
-  if (length(by_group) <= 1 & !is.null(company_aggregated_alignment_net)) {
+  if (!is.null(company_aggregated_alignment_net)) {
     data_sankey_sector <- prep_sankey(
       company_aggregated_alignment_net,
       region = "global",
@@ -174,7 +174,7 @@ plot_aggregate_loanbooks <- function(config) {
     print("Sankey plot (by sector) cannot be generated. Skipping!")
   }
 
-  if (length(by_group) <= 1 & !is.null(company_aggregated_alignment_net)) {
+  if (!is.null(company_aggregated_alignment_net)) {
     data_sankey_company_sector <- prep_sankey(
       company_aggregated_alignment_net,
       region = "global",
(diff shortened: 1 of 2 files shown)


jdhoffa:
Is it correct that this behaviour was introduced in #179?

R/plot_aggregate_loanbooks.R
```diff
@@ -63,69 +63,85 @@ plot_aggregate_loanbooks <- function(config) {
...
+  if (file.exists(file.path(output_analysis_aggregated_dir, glue::glue("company_exposure_net_aggregate_alignment{file_by_group}.csv")))) {
+    company_aggregated_alignment_net  <-
+      readr::read_csv(
+        file = file.path(
+          output_analysis_aggregated_dir,
+          glue::glue("company_exposure_net_aggregate_alignment{file_by_group}.csv")
+        ),
+        col_types = col_types_company_aggregated_alignment,
+        col_select = dplyr::all_of(col_select_company_aggregated_alignment)
+      )
```
cjyetman: I'd rather see this filename "glued" together once above, then tested for existence, and then possibly read from.

R/plot_aggregate_loanbooks.R
```diff
@@ -63,69 +63,85 @@ plot_aggregate_loanbooks <- function(config) {
...
-      col_types = col_types_company_aggregated_alignment,
-      col_select = dplyr::all_of(col_select_company_aggregated_alignment)
-    )
+  if (file.exists(file.path(output_analysis_aggregated_dir, glue::glue("company_exposure_bo_po_aggregate_alignment{file_by_group}.csv")))) {
+    company_aggregated_alignment_bo_po <-
+      readr::read_csv(
+        file = file.path(
+          output_analysis_aggregated_dir,
+          glue::glue("company_exposure_bo_po_aggregate_alignment{file_by_group}.csv")
+        ),
```
cjyetman: same here

cjyetman: and so forth below

jacobvjk:
> Is it correct that this behavior was introduced in #179?

Partially. Prior to #179 and #177 `analyse()` would have errored out ungracefully in `run_aggregate_alignment_metric()`. Post these two PRs the initial source had been fixed, but there was still the possibility to error out elsewhere, which is addressed here. So this patch complements the other two PRs. They did not break anything that was not already broken. Now with better testing, we might have caught the need to add these checks earlier, but the whole point of #177 is to enable useful integration tests in the first place

R/plot_aggregate_loanbooks.R
```diff
@@ -218,56 +236,58 @@ plot_aggregate_loanbooks <- function(config) {
...
-    ) {
-      data_scatter_alignment_exposure <- loanbook_exposure_aggregated_alignment_net %>%
-        prep_scatter_alignment_exposure(
-          year = year_scatter_alignment_exposure,
-          region = region_scatter_alignment_exposure,
-          scenario = scenario_select,
-          group_var = by_group,
-          exclude_groups = "benchmark"
-        )
+    if (!is.null(loanbook_exposure_aggregated_alignment_net)) {
```
cjyetman: can these multi-level if statements be combined with `&` or `|`? seems overly complicated at the moment

cjyetman:
> Is it correct that this behaviour was introduced in #179?

I don't think so

R/plot_aggregate_loanbooks.R
jacobvjk: one would hope so, not totally sure how best to do it though. I am not sure testing for `is.null(df)` and `nrow(df) > 0` can be combined? At least I thought they could not. But to avoid gross repetition, I could define the conditions further above and re-use them where possible. Also I guess `length(by_group)` could possibly be combined with `nrow(df)`... will check

R/plot_aggregate_loanbooks.R
jacobvjk: done

R/plot_aggregate_loanbooks.R
jacobvjk: done

R/plot_aggregate_loanbooks.R
jacobvjk:
```
df <- NULL

if (!is.null(df)) {
  if (
    nrow(df) > 0
  ) {
    print("df is not empty")
  } else {
    print("df is empty")
  }
} else {
  print("df is NULL")
}
#> [1] "df is NULL"

df <- tibble::tibble()

if (!is.null(df)) {
  if (
    nrow(df) > 0
  ) {
    print("df is not empty")
  } else {
    print("df is empty")
  }
} else {
  print("df is NULL")
}
#> [1] "df is empty"

df <- tibble::tibble(a = 1)

if (!is.null(df)) {
  if (
    nrow(df) > 0
  ) {
    print("df is not empty")
  } else {
    print("df is empty")
  }
} else {
  print("df is NULL")
}
#> [1] "df is not empty"

df <- NULL

!is.null(df) & nrow(df) > 0
#> logical(0)

if (!is.null(df) & nrow(df) > 0) {
  print("does this work?")
} else {
  print("does this work?")
}
#> Error in if (!is.null(df) & nrow(df) > 0) {: argument is of length zero
```

Created on 2024-10-30 with [reprex v2.1.0](https://reprex.tidyverse.org/)

cjyetman (approved):

PR #183, opened 2024-10-31
fix reference to filepath variable

cjyetman:
is this correct @jacobvjk?

### DIFF
--- a/R/prepare_abcd.R
+++ b/R/prepare_abcd.R
@@ -26,7 +26,7 @@ prepare_abcd <- function(config) {
   stop_if_not_inherits(output_prepare_dir, "character")
 
   if (dir.exists(output_prepare_dir)) {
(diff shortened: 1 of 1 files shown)


jacobvjk:
I meant to indicate the name of the parameter from the config file. `output_prepare_dir` is not a variable visible to the user. But it may be better to just paste the actual path.. what's your view on that?

cjyetman:
> I meant to indicate the name of the parameter from the config file. `output_prepare_dir` is not a variable visible to the user. But it may be better to just paste the actual path.. what's your view on that?

got it.... "name of the parameter from the config file" makes more sense

jacobvjk (commented):
see comment above. I would not make this change, but maybe we want to display the actual path instead?

cjyetman:
> see comment above. I would not make this change, but maybe we want to display the actual path instead?

maybe the path, but I think for now the key name in the config is appropriate and in line with messages we show to users in other places

PR #193, opened 2024-11-04
copy needed files for tests into temp dir and remove after tests

### LINKED ISSUE #188: find a way that allows integration testing without requiring manual clean up of test dirs for local testing
> running the tests locally seems a bit brittle. when I run them, I get this:
> 
> ```
> ✔ |          1 | 3-prioritise_and_diagnose [2.1s]            
> ⠏ |          0 | 4-analyse                                   [1] "Sankey plot (by sector) cannot be generated. Skipping!"
> [1] "Sankey plot (by sector and company) cannot be generated. Skipping!"
> [1] "Scatter plot exposure by alignment cannot be generated. Skipping!"
> Scatter plot BO/PO cannot be generated. Skipping!
> Scatter plot BO/PO cannot be generated. Skipping!
> Scatter plot BO/PO cannot be generated at company level. Skipping!
> Scatter plot BO/PO cannot be generated at company level. Skipping!
> ✔ |          1 | 4-analyse 
> ```
> 
> is that expected? It seems like `4-analyse` is run twice, once skipping, once successfully? Also, when I now rerun the test, I get an error in `1-prepare_abcd`, because the output directory already exists (i.e. it was not cleaned up after the first run) and the test expects no conditions. Although the condition is just the warning that the output directory already exists and will be overwritten, as per #179
> 
> ```
> ── Failed tests ─────────────────────────────────────────────
> Failure (test-1-prepare_abcd.R:28:3): with known input, runs without error
> Expected `prepare_abcd(config)` to run without any conditions.
> i Actually got a <simpleWarning> with text:
>   Output directory `dir_prepared_abcd` already exists. The existing directory will be removed and replaced with the output of the current run.
> ```

1 - yes, it is expected that those "sanky" lines get output to the console... it then forces testthat to reprint the test name line so that it is the most recent thing on the console, it's *not* actually running again

2 - I couldn't think of a clean way of clearing out the test files that are created during the test because they need to be intermingled with the files that need to be created pre-testing which are included in the repo. In the automated CI, it doesn't matter because a brand new clone is always made first that will always have the pre-files but not the post-files. I would dump them in a temp directory that gets deleted when the test are done, but they need to be intermingled with the input files that are already there. I can't make them use temporary files because we're using functions that have a side effect of creating files. catch-22

_Originally posted by @cjyetman in https://github.com/RMI-PACTA/pacta.multi.loanbook/issues/185#issuecomment-2451720594_
            

AB#12075

cjyetman:
- closes #188

### DIFF
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -1,0 +1,7 @@
+# Run before any test
+file.copy(test_path("test-data"), tempdir(), recursive = TRUE)
+test_tmpdir <- file.path(tempdir(), "test-data")
+
+
+# Run after all tests
+withr::defer(unlink(test_tmpdir), teardown_env())
--- a/tests/testthat/test-1-prepare_abcd.R
+++ b/tests/testthat/test-1-prepare_abcd.R
@@ -2,8 +2,8 @@ test_that("with known input, runs without error", {
   config <-
     list(
       directories = list(
-        dir_input = test_path("test-data", "input"),
-        dir_prepared_abcd = test_path("test-data", "output", "prepared_abcd")
+        dir_input = file.path(test_tmpdir, "input"),
+        dir_prepared_abcd = file.path(test_tmpdir, "output", "prepared_abcd")
       ),
       file_names = list(
         filename_abcd = "ABCD.xlsx",
--- a/tests/testthat/test-2-match_loanbooks.R
+++ b/tests/testthat/test-2-match_loanbooks.R
@@ -2,9 +2,9 @@ test_that("with known input, runs without error", {
   config <-
     list(
       directories = list(
-        dir_input = test_path("test-data", "input"),
-        dir_prepared_abcd = test_path("test-data", "output", "prepared_abcd"),
-        dir_matched_loanbooks = test_path("test-data", "output", "matched_loanbooks")
+        dir_input = file.path(test_tmpdir, "input"),
+        dir_prepared_abcd = file.path(test_tmpdir, "output", "prepared_abcd"),
+        dir_matched_loanbooks = file.path(test_tmpdir, "output", "matched_loanbooks")
       ),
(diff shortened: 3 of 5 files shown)


jdhoffa (approved):
woop woop!!

jacobvjk (approved):
this works for me if I use `devtools::check` multiple times and if I use `devtools::test()` restarting the R session between runs. It weirdly does not work for me with `devtools::test()` without restarting the R session. Not a blocker for me, but interesting. But since `devtools::test()` is the same as `testthat::test_dir()` and `test_dir()` documentation says the following:

```
This function is the low-level workhorse that powers [test_local()](http://127.0.0.1:11373/help/library/testthat/help/test_local) and [test_package()](http://127.0.0.1:11373/help/library/testthat/help/test_package). Generally, you should not call this function directly.
```
... I guess I am fine with that

cjyetman:
> this works for me if I use `devtools::check` multiple times and if I use `devtools::test()` restarting the R session between runs. It weirdly does not work for me with `devtools::test()` without restarting the R session. Not a blocker for me, but interesting. But since `devtools::test()` is the same as `testthat::test_dir()` and `test_dir()` documentation says the following:
> 
> ```
> This function is the low-level workhorse that powers [test_local()](http://127.0.0.1:11373/help/library/testthat/help/test_local) and [test_package()](http://127.0.0.1:11373/help/library/testthat/help/test_package). Generally, you should not call this function directly.
> ```
> 
> ... I guess I am fine with that

I think that `devtools::test()`, and effectively `testthat::test_dir()`, will not run the teardown commands properly when run interactively, so the temp directory does not get cleaned up. I always use the Test button in the Build pane, which runs in a clean environment. I suspect that restarting R works because it sets a new temp directory, ignoring the previous one that is still there.

PR #194, opened 2024-11-04
convert print commands to proper warnings

cjyetman:
This is better practice and will allow for squelching this output when running tests.


jdhoffa (approved):
looking good!

jacobvjk (approved):

PR #203, opened 2024-11-07
add manifest output

cjyetman:
outputs a manifest YML at the end of each of the 4 main steps, which includes the current config being used and the directory and timestamp of the manifests from the output of the prior steps used


R/match_loanbooks.R
```diff
@@ -170,5 +170,12 @@ match_loanbooks <- function(config) {
       )
     cli::cli_progress_update()
   }
+  
   cli::cli_progress_done()
+  
+  write_manifest(
+    config = config,
+    path = file.path(output_matched_loanbooks_dir, "manifest.yml"),
+    prior_input_paths = c(output_prepare_dir)
```
jacobvjk, line 179: why not use `get_output_prepare_dir(config)` here, as is done for `analyse()`

R/analyse.R
```diff
@@ -35,6 +35,16 @@ analyse <- function(config) {
   run_pacta(config)
   run_aggregate_alignment_metric(config)
   plot_aggregate_loanbooks(config)
+  
+  write_manifest(
+    config = config,
+    path = file.path(output_analysis_dir, "manifest.yml"),
+    prior_input_paths = c(
+      get_output_prepare_dir(config),
+      get_output_matched_loanbooks_dir(config),
```
jacobvjk, line 44: `analyse()` does not technically load `matched_loanbooks` for any of the functions it calls. I am a bit on the fence on whether that should mean that we do not add the timestamp of `matched_loanbooks` (no direct dependency and you can backtrace the timestamp by checking `prioritized_loanbooks_and_diagnostics`, which does depend on `matched_loanbooks`) or we do add it (because it is impossible to run `analyse()` without having previously run `match_loanbooks()`...

R/write_manifest.R
```diff
@@ -0,0 +1,46 @@
+write_manifest <- function(config, path, prior_input_paths = NULL) {
+  timestamp <-
+    as.character(
+      format(
+        x = Sys.time(),
+        format = "%Y-%m-%d %H:%M:%OS6",
+        tz = "UTC"
```
jacobvjk: if we enforce UTC time zone, we should probably document this somewhere

jacobvjk (commented):
works, but see comments

R/match_loanbooks.R
cjyetman: I questioned this too, but the output path, where the manifest should go, for any given function is already assigned to a variable and the function already used above, in this case `get_output_prepare_dir(config)`, so it seemed appropriate to use it rather than run the same function again?

cjyetman: and I guess I wanted to avoid some future possibility that the path had been modified in the function

R/analyse.R
cjyetman: also on the fence, so I went for more information instead of less?

R/match_loanbooks.R
jacobvjk: yea I guess. `match_loanbooks()` is a bit different in that it does not only call some lower level functions, but goes through the full process directly.. so I guess slightly different handling is appropriate

R/write_manifest.R
cjyetman: would it make more sense to use the local timezone? I think that is achievable

jacobvjk: might be better, assuming the system this runs on actually has that set up appropriately. I remember an application with a supervisor last year, where the system was actually set to UTC although that was far off their actual timezone. somewhat confusing. I guess we could just use `Sys.time()` and also write `Sys.timezone()`?? as long as it is clearly spelled out, I am fine with either UTC or local

cjyetman: If we leave out the `tz` specification entirely it defaults to `""` which is effectively the current "local" timezone. If a user's local timezone is set to UTC even if that's not geographically correct for them I think it is their problem.

cjyetman: removed the explicit timezone specification

jacobvjk: okay, works for me

R/analyse.R
jacobvjk: let's keep it for now, probably no reason not to

jacobvjk (approved):

PR #204, opened 2024-11-07
remove unsued `abcd_test_data` and `loanbook_test_data`

### LINKED ISSUE #118: reconsider if `abcd_test_data` and `loanbook_test_data` should be exported
maybe also `data_dictionary` if its primary purpose is to drive [the article/vignette page](https://rmi-pacta.github.io/workflow.multi.loanbook/articles/data_dictionary.html) and it's not intended to be accessed directly by a user?

AB#12145

cjyetman:
- closes #118


jacobvjk (approved):
lgtm

PR #205, opened 2024-11-07
remove `sector_split_type` from output directory of analysis

jacobvjk:
closes ADO https://dev.azure.com/RMI-PACTA/2DegreesInvesting/_workitems/edit/12181

- removes a level of the analysis output directory, that was previously inserted IF the sector split was applied. The name of the directory was the type of sector split, so effectively it was always `equal_weights` or nothing at all
- The information on whether or not a sector split was applied and what type of sector split can be obtained from the config as well and from the manifest files (see #203 )

Previously, the output directory for the analysis would have looked something like this for when the sector split was applied

```
project_folder
├── config.yml
└── analysis_aggregate
    └── equal_weights
        ├── standard
        └── aggregated

```
and like this when it was not applied:

```
project_folder
├── config.yml
└── analysis_aggregate
    ├── standard
    └── aggregated

```

Now, the output directory always follows the latter structure


cjyetman (approved):
makes sense but I haven't actually tested this

PR #211, opened 2024-11-08
docs: streamline `README.md`

jdhoffa:
In this PR, I try to make the `README.md` much less verbose. 

[The cookbook](https://rmi-pacta.github.io/pacta.multi.loanbook/articles/cookbook.html) already contains very expansive and detailed technical information with everything you could want to know. 

The goal of the `README.md` should be to provide JUST enough information for the user to get things started, without being daunting.


jdhoffa:
It also makes it marginally easier to maintain long-term, in case sections etc. in the cookbook change

jacobvjk (approved):

PR #223, opened 2024-11-11
don't import `remove_inactive_companies` when not needed

### LINKED ISSUE #206: only get the parameter values that are used in each step
`run_pacta()` and `plot_aggregate_loanbooks()` both get `remove_inactive_companies`, but neither one of them uses it

cjyetman:
- closes #206


jacobvjk (approved):
nice one

PR #224, opened 2024-11-12
automatically create dev and release versions of the pkgdown site

### LINKED ISSUE #151: consider having a `release` and `dev` version of the pkgdown website
As the r2dii packages do, it's possible to automatically maintain two pkgdown websites, one for the "release" version (generated at the root level of the pkgdown website whenever a standard semantic version number is used in the DESCRIPTION) and a "dev" version for whatever is on main whenever a version number has four components, e.g. `0.0.0.9000`

how-to documented here: https://pkgdown.r-lib.org/reference/build_site.html#development-mode

cjyetman:
- closes #151 

As I understand it, setting the mode to "auto" ("release" is default), whenever a push to main is made the version number will be checked and if it has four components, e.g. 0.0.0.9000, it will be considered a dev version and create a dev site under `doc/dev/`, otherwise it will create a "release" site under `doc/`. A release site will not overwrite the current dev site and vice versa.

In the current state where there is no release or dev version and the version number has four components, it will create an "unreleased" site at `doc/` with the version number in red and with a pop-up page that says "unreleased". I tested this locally and it seems that's how it behaves. Presumably, as soon as we have a release version, the behavior will automatically follow the above logic.

official documentation here: https://pkgdown.r-lib.org/reference/build_site.html#development-mode


jdhoffa (approved):
Indeed! This is how it's done :-)

(Confirmed it's the same implementation here: https://github.com/RMI-PACTA/r2dii.analysis/blob/2f606b083ec9db9679fb9c94823e87d655ee8bb7/_pkgdown.yml#L2-L3 )

jacobvjk (approved):

Issue #210, opened 2024-11-08
Once on CRAN, update `README.md` with CRAN installation instructions


jdhoffa:
Currently we only explain how to install development version. 

Depends on package being accepted to CRAN.

jdhoffa:
Also update the Cookbook installation instructions. 
The Prod site should always suggest using `CRAN`
The dev site can also suggest using `pak`

jdhoffa:
Now that the package is on CRAN, I think we should update installation instructions to prefer `install.packages()` 
FYI @cjyetman @jacobvjk

cjyetman:
Currently, the README and the Preparatory Steps page of the Cookbook include instructions on how to install with `install.packages()` (from CRAN) *and* how to install the dev version with `pak::pak()`, so I think this can be closed

PR #235, opened 2024-11-13
add different install instructions for release and dev

### LINKED ISSUE #217: docs: Remove CRAN installation instructions until it is actually on CRAN
We anticipate that it will be accepted to CRAN, but if it is not then these instructions will only be confusing (especially if we forget about them). 

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/c0bc24e76f7f7b72d4c3e0bf9458011f0d4f4aa1/vignettes/cookbook.Rmd#L210-L214

Relates to #210

cjyetman:
- closes #217 
- sets us up to do #210 very easily once it's released on CRAN (by simply removing the comment fences)

note: also removes the text saying which repo `install.packages()` installs from, which technically can not be know since it's dependent on the user's setup

note: also removes the text about verifying the install with `library(pacta.multi.loanbook)` because I don't think that makes much sense nor is it really a great way to actually verify install

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 [![Lifecycle: experimental](https://img.shields.io/badge/lifecycle-experimental-orange.svg)](https://lifecycle.r-lib.org/articles/stages.html#experimental)
 [![R-CMD-check](https://github.com/RMI-PACTA/pacta.multi.loanbook/actions/workflows/R-CMD-check.yaml/badge.svg)](https://github.com/RMI-PACTA/pacta.multi.loanbook/actions/workflows/R-CMD-check.yaml)
 [![CRAN
-status](https://www.r-pkg.org/badges/version/pacta.multi.loanbook)](https://CRAN.R-project.org/package=pacta.multi.loanbook)
+status](https://www.r-pkg.org/badges/version/pacta.multi.loanbook)](https://CRAN.R-project.org/package=pacta.multi.loanbook){.pkgdown-release}
 [![pacta.multi.loanbook status badge](https://rmi-pacta.r-universe.dev/badges/pacta.multi.loanbook)](https://rmi-pacta.r-universe.dev/pacta.multi.loanbook)
 <!-- badges: end -->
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -5,7 +5,7 @@
 [![Lifecycle: experimental](https://img.shields.io/badge/lifecycle-experimental-orange.svg)](https://lifecycle.r-lib.org/articles/stages.html#experimental)
 [![R-CMD-check](https://github.com/RMI-PACTA/pacta.multi.loanbook/actions/workflows/R-CMD-check.yaml/badge.svg)](https://github.com/RMI-PACTA/pacta.multi.loanbook/actions/workflows/R-CMD-check.yaml)
 [![CRAN
-status](https://www.r-pkg.org/badges/version/pacta.multi.loanbook)](https://CRAN.R-project.org/package=pacta.multi.loanbook)
+status](https://www.r-pkg.org/badges/version/pacta.multi.loanbook)](https://CRAN.R-project.org/package=pacta.multi.loanbook){.pkgdown-release}
```
jdhoffa: This seems to not render correctly:
Screenshot 2024-11-13 at 19 19 24

jdhoffa (changes requested):
It's hard for me to test if the `::: {.pkgdown-release}`  chunks render without seeing it on `pkgdown`, but I'll take your word!

The `README.md` does not seem to render as expected though.

README.md
jdhoffa: @cjyetman pretty sure this needs to be a `README.Rmd` for that to work
😂 your favourite!!

cjyetman: > @cjyetman pretty sure this needs to be a `README.Rmd` for that to work 😂 your favourite!!

hmmm, you might be right... it works on the main page of the pkgdown site, but maybe/probably will be ignored on GitHub's rendering of the README.md

I'll just comment fence it for now

cjyetman:
> It's hard for me to test if the `::: {.pkgdown-release}` chunks render without seeing it on `pkgdown`, but I'll take your word!
> 
> The `README.md` does not seem to render as expected though.

one can run `pkgdown::build_site()` to see the result locally

README.md
cjyetman: commented out

jdhoffa (approved):
🚀

jacobvjk (approved):

PR #240, opened 2024-11-18
return last company information as output file, not as warning

### LINKED ISSUE #74: improve warning about change in unique companies after sector split
Currently when I run `run_match_prioritize()`, and more specifically `apply_sector_split_to_loans()`, with the test loanbooks and data that I have, I get all of these warnings. To be honest, I'm not sure if that's bad (probably, because it's a warning?), I'm not sure *how bad* that is (e.g. going from 1 to 132 seems dramatic, but maybe that's normal?), and I don't have any idea of how to fix it or what to look at. It would be advantageous to improve this a bit. Additionally, I suspect if this happens to 1 loanbook it's likely to happen in others (?), so when one runs `run_match_prioritize()` they'll likely get a series of these warnings one after another, and it's not clear which loanbook a given warning message is related to, so maybe we could improve that?

``` r
run_match_prioritize(config_path)
#> Warning messages:                                                                             
#> 1: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 74 unique companies. 
#> 2: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 84 unique companies. 
#> 3: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 51 unique companies. 
#> 4: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 76 unique companies. 
#> 5: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 115 unique companies. 
#> 6: ! Applying the sector split has lead to changes in the number of unique companies covered in
#>   the analysis.
#> ℹ Prior to the split, there was 1 unique company.
#> ℹ After the split, there are 132 unique companies. 
```

jacobvjk:
closes #74 

- instead of warning the users about companies lost in the sector_split, outputs are now generated that document the lost companies for later inspection
- data dictionary is updated accordingly with information about the new output file


cjyetman (approved):
Makes sense and works as expected as far as I can tell locally. My only minor concern is that rather than spitting out a warning to the console for every loanbook included in the process (which could be 1000's maybe), now it will create a new CSV in the `prioritized_loanbooks_and_diagnostics` output directory for every loanbook included in the process? Maybe in the future we can think about how to better organize the `prioritized_loanbooks_and_diagnostics` output directory.

jacobvjk:
> Makes sense and works as expected as far as I can tell locally. My only minor concern is that rather than spitting out a warning to the console for every loanbook included in the process (which could be 1000's maybe), now it will create a new CSV in the `prioritized_loanbooks_and_diagnostics` output directory for every loanbook included in the process? Maybe in the future we can think about how to better organize the `prioritized_loanbooks_and_diagnostics` output directory.

Yes, that is a fair concern, structurally. For the moment I think this is fine, because the structure of the outputs will also be one file for each prioritized loan book, so there is some correspondence there. Also, the cases we have seen so far have always been significiantly below a case number of thousands of loan books. But I agree that it could be better nonetheless to gather all this information and write it into one file per run.

Issue #250, opened 2024-11-19
rework output format of `lbk_match_success_rate`


jacobvjk:
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/2a2a706249b41b4233d093f626e6719126b8e6e8/data-raw/data_dictionary.R#L256

the output table is not formatted exactly identical to the input table used for the match success rate plot. This could be confusing.

relates to #213

Issue #249, opened 2024-11-19
consider deprecating output `companies_included`


jacobvjk:
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/2a2a706249b41b4233d093f626e6719126b8e6e8/data-raw/data_dictionary.R#L230

There is no obvious use case for this output. It does not confuse much, but it also may not be helpful

relates to #213

Issue #248, opened 2024-11-19
rework output type `data_scatter_sector`


jacobvjk:
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/2a2a706249b41b4233d093f626e6719126b8e6e8/data-raw/data_dictionary.R#L147

relates to #213 

although the relevant output file only contains information for one sector at a time, the table does not have a sector column. The sector is only indicated in the file name. It would be better to indicate in the actual data what sector the result refers to

Issue #213, opened 2024-11-08
upkeep: Review and remove all `TODO`, `FIXME`, and any other comments used to kick cans down roads


jdhoffa:
"Remove" here means: 
* The comment is reviewed
* An assessment is made if that comment is still relevant
* If it IS relevant, a GitHub issue is created explaining what needs to happen
* If it IS NOT relevant, the comment is deleted

---

[search for `TODO` in the code](https://github.com/search?q=repo%3ARMI-PACTA%2Fpacta.multi.loanbook+todo&type=code)
[search for `**LINK**` in the code](https://github.com/search?q=repo%3ARMI-PACTA%2Fpacta.multi.loanbook+**link**&type=code)

jdhoffa:
Do this also for missing `LINK`s:
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/c0bc24e76f7f7b72d4c3e0bf9458011f0d4f4aa1/vignettes/cookbook.Rmd#L106

cjyetman:
I think many of the "todo"s meant to prompt creating examples should simply be removed, especially those attached to unexported functions like the plotting functions.

cjyetman:
I've reviewed all of these and there aren't an easy wins... these will all take some consideration from @jacobvjk I believe.

jdhoffa:
That's ok! I didn't necessarily expect them to be easy wins (in terms of actually fixing them). 

The main goal of this issue is that we migrate them all into well-specified GitHub issues. A `TODO` (et.c) is likely to just get left/ forgotten about, context gets lost, and they just end up as tech debt hanging around the code

jdhoffa:
Or did you mean that you also don't have the context to extract issues?
In that case, agreed probably this is a job for @jacobvjk

cjyetman:
I just thought a bunch of them would be e.g. "todo" examples of functions that we are no longer exporting, but I think only 4 of them are like that... the rest I don't have enough context to know if they're still relevant.

jdhoffa:
Got it ok. Let's triage this next sprint then. Likely a @jacobvjk ticket

cjyetman:
and actually, all of the todo-examples-of-using-no-longer-exported-plotting-functions I think were already stripped... the remaining todos in the plotting functions are things that need more context than I have also.

PR #251, opened 2024-11-19
Pick up cans down the road

jacobvjk:
relates to #213 

- removes obsolete TODOs
  - `# TODO: allow surfacing the other match_name args` --> no plans to do so
  - `# TODO: when benchmarks get re-introduced, they need to be removed here` --> no plans to reintroduce benchmarks for net alignment metric. In case we want to do so in the future, we should reconsider the implementation more broadly
  - `# TODO: Note that this implies that no groups across different .by variables
  should have the same values, as this will confuse output directories` --> this was effectively handled by https://github.com/RMI-PACTA/pacta.multi.loanbook/pull/179, because every additional run with a new by_group variable is either going to be written to a new directory (hence there will be no confusion of equally named sub-directories), or it will replace the previous output directory entirely (hence again no naming conflicts for equally named sub groups created in different runs)
  - `# TODO: check if this needs to be adjusted to remove other by_group columns` the extended implementation of `by_group` (#108, #120, #127) has not lead to any issues with generating outputs
  - `# TODO: decide if this should be removed from outputs` --> no pressing need to remove a column that indicates the calculation was run at meta level
- relevant TODOs have received their own GH issues (#248, #249, #250)
- close TODOs related to config validations by adding the relevant assertions

Note

TODOs related to the plots, tests, and examples in the user-facing functions will be handled in another PR

### DIFF
--- a/R/match_loanbooks.R
+++ b/R/match_loanbooks.R
@@ -79,10 +79,16 @@ match_loanbooks <- function(config) {
     assert_inherits(matching_p, "numeric")
   }
 
-  # TODO: check for data.frame
-  # assert_length(matching_overwrite, 1L)
-  # assert_inherits(matching_overwrite, "numeric")
-  #
+  if (!is.null(matching_overwrite)) {
+    assert_inherits(matching_overwrite, "data.frame")
+    # cols are based on r2dii.data::overwrite_demo
+    assert_expected_columns(
+      data = matching_overwrite,
+      cols = c("level", "id_2dii", "name", "sector", "source"),
+      desc = "matching_overwrite"
+    )
+  }
+
   # TODO: check for join_object
   # assert_length(matching_join_id, 1L)
   # assert_inherits(matching_join_id, "numeric")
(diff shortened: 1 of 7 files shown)


cjyetman (approved):
all looks reasonable to me, though to be honest there's a lot going on here so it's hard for me to fully tell if some things were properly addressed or not, e.g. this comment was removed "# TODO: when benchmarks get re-introduced, they need to be removed here", but it's not clear to me if/where that is be handled elsewhere in this PR

jacobvjk:
> all looks reasonable to me, though to be honest there's a lot going on here so it's hard for me to fully tell if some things were properly addressed or not, e.g. this comment was removed "# TODO: when benchmarks get re-introduced, they need to be removed here", but it's not clear to me if/where that is be handled elsewhere in this PR

yes, I realize that if we want to track all of these properly, I would have to make one Pr for each type of TODO. the one you mentioned was removed as obsolete because there is no plan on the horizon to reintroduce corporate economy benchmarks for the net alignment metric

jacobvjk:
I added a more detailed description to the PR to ensure the reasoning for every removal can be tracked

PR #255, opened 2024-11-20
split cook book in sections for each chapter

### LINKED ISSUE #246: docs: split cookbook by section
See how the "Documentation" header has different sections, however the "Cookbooks" header only has the "Cookbook" section. 
We should split it into the top-level sections. 

Relates to #215 

Screenshot 2024-11-19 at 11 54 13

jacobvjk:
closes #246 

- gains five separate chapters below the Cookbook article entry
- all five sections are linked to the previous/next chapter at the bottom


jacobvjk:
it seems like the file size of the vignettes does not go down well with the R-CMD check and triggers a note. @cjyetman  @jdhoffa what do you think about making the cook book a set of articles instead of vignettes?

https://r-pkgs.org/vignettes.html#sec-vignettes-article

cjyetman:
I'm not opposed to that.

jacobvjk:
I am not sure the figures render.. locally I seem to struggle with that.. but I can't see why they shouldn't


`Missing images in 'vignettes/cookbook_running_the_analysis.Rmd': '../reference/figures/p4s_workflow_structure.png' ℹ pkgdown can only use images in 'man/figures' and 'vignettes'`

jdhoffa:
> it seems like the file size of the vignettes does not go down well with the R-CMD check and triggers a note. @cjyetman @jdhoffa what do you think about making the cook book a set of articles instead of vignettes?
> 
> https://r-pkgs.org/vignettes.html#sec-vignettes-article

Indeed, I think this is actually preferable/ what is recommended by pkgdown etc.
Shall we tackle that in a separate issue/ PR?

jacobvjk:
> > it seems like the file size of the vignettes does not go down well with the R-CMD check and triggers a note. @cjyetman @jdhoffa what do you think about making the cook book a set of articles instead of vignettes?
> > https://r-pkgs.org/vignettes.html#sec-vignettes-article
> 
> Indeed, I think this is actually preferable/ what is recommended by pkgdown etc.

I refrained from that for the time being, because rendering the articles errored out. pkgdown did not manage to find the figures and I am unsure why

cjyetman (approved):

PR #258, opened 2024-11-21
add `initialize_default_project()`

### LINKED ISSUE #247: feat: gains function `initialize_default_project`
User Story: 
I am a new user of `pacta.multi.loanbook` and I would like to do a full dry run of the analysis with sane defaults. 

Acceptance Criteria: 

- [ ] `example.config.yml` should be removed from the root of the GH repo, it should only exist in the `inst` folder as part of the package
- [ ] function should take an argument to specify the desired new directory. If it already exists, error that you can't overwrite an existing project 
- [ ] If no argument is specified, default to writing a `project` directory to the current working directory
- [ ] function should write a `config.yml` that is similar to the default `example.config.yml`, however with all input/output directories appropriately specified
- [ ] all other values should be pre-filled with sane defaults
- [ ] all sub-directories should be created with sane default names
- [ ] `initialize_default_project` should be based off of this initilization function

Relates to and supersedes #219

cjyetman:
- closes #247 

I followed the specifications set in #247 precisely, but honestly I don't think this works well... for one, it creates the sub-directories, but those sub-directories are created by the functions that run each of those processes, so if you use `initialize_default_project()` you'll get a warning about overwriting an existing directory in each of the following steps. I also feel like it is somewhat at odds with the manifest files we're creating after each step of the process because the config.yml for each step of the process could/should be progressively modified, which implies there should not be a single config file.

⚠️ in order to write a YAML file that looks like the example (and our documentation) I had to use multiple "handlers" for `yaml::as.yaml()`, otherwise strings get unquoted, logicals show up as `true/false`, and `NULL` shows up as `~` 🤷🏻


jdhoffa:
@cjyetman can you suggest what you think a more optimal implementation would be on the original issue please?

The AC there were meant as a starting point, very happy to amend them if they suggested something sub-optimal

I agree that having warnings bubble up from the suggested default workflow is silly haha 😂

jacobvjk:
Maybe it is sufficient to create the input sub-directory, which would not be created by any other function. Otherwise we can decide that no sub-directories should be created and clarify to the user that they need to create the input sub-diorectory manually and in accordance to the path set in the config.yml. For all output directories, setting the directories in the config would be sufficient to have them created by the following function calls without generating warnings.

R/initialize_default_project.R
```diff
@@ -0,0 +1,57 @@
+#' Initialize a project directory and its config file
+#'
+#' @param path an absolute or relative path pointing to the location you would
+#'   like the project directory to be created
+#'
+#' @returns NULL
+#'
+#' @export
+
+initialize_default_project <- function(path = "project") {
```
jacobvjk: Personally, I would prefer to call the function `initialise_default_project()` and add somehting like 

```
#' @rdname initialise_default_project
#' @export
initialize_default_project <- initialise_default_project
```

This keeps the naming across the repository consistent.

jacobvjk (changes requested):
one specific change and once we settle on the exact implementation, we will need to look at changes in the cookbook

cjyetman:
> Maybe it is sufficient to create the input sub-directory, which would not be created by any other function. Otherwise we can decide that no sub-directories should be created and clarify to the user that they need to create the input sub-diorectory manually and in accordance to the path set in the config.yml. For all output directories, setting the directories in the config would be sufficient to have them created by the following function calls without generating warnings.

Yeah, I think it makes more sense to only make the input directory, the rest are defined by the config and the functions will create them if necessary (or warn about overwriting).

I still feel some tension about the config that I can't quite explain. I guess our cookbook suggests that one might/should run parts of the process multiple times with tweaked settings, which implies to me that they may be using multiple config files and have multiple roughly equivalent sub directories, e.g. `analyse_by_group` and `analyse_by_sector`. And we even facilitated using full paths to each directory, so those sub directories don't even need to be in the same project directory (iirc we specifically thought that the `input` directory might likely be in a separate location).

I don't have a good suggestion here (sorry), just being forward about a bit of hesitation I have.

R/initialize_default_project.R
cjyetman: thanks for keeping my American English in check 😁

jacobvjk: I will admit that working for an American organization, the choice is questionable. But it is what it is at this point

cjyetman: done

cjyetman:
Did some significant reworking to only create the project and input directories, and then deal with getting absolute paths for the other directories that don't yet exist.

jacobvjk (approved):
I think this should work pretty well now. We need a follow up documentation task for the cookbook, but in principle this is a good way to set up a project

PR #262, opened 2024-11-22
docs: remove explicit cookbook section numbering

jdhoffa:
The order is defined implicitly by the order in which it appears on the `pkgdown.yml`:
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/3f08f0ded38803dd34937155eb16cfece32e4a30/_pkgdown.yml#L41-L46

Explicit numbering of the sections is superfluous and clunky. 
Screenshot 2024-11-22 at 13 02 13

### DIFF
--- a/vignettes/cookbook_overview.Rmd
+++ b/vignettes/cookbook_overview.Rmd
@@ -49,6 +49,14 @@ plot_table <- function(table) {
 
 This cookbook provides a step-by-step guide to running the PACTA for Supervisors analysis using the `pacta.multi.loanbook` package. The analysis is designed to help financial supervisors assess the alignment of banks' loan books with the Paris Agreement goals.
 
+The sections of the cookbook are as follows:
+
+1. **Overview**: This section provides an overview of the PACTA for Supervisors analysis and its main steps.
+2. **Preparatory Steps**: Prepare the ABCD data and optionally a custom sector split.
+3. **Running the Analysis**: Match the raw loan books to the ABCD data and run the PACTA for Supervisors analysis.
+4. **Interpretation of Results**: Interpret the results of the analysis and understand the outputs.
+5. **Advanced Topics**: Learn about advanced topics such as custom grouping of results and how to adjust the analysis parameters for best results on advanced research questions.
+
 ## What is the PACTA for Supervisors analysis?
 
 PACTA for Supervisors is based on the PACTA methodology, which assesses the alignment of financial portfolios with climate goals utilizing forward-looking asset-based company data (ABCD) that is linked to financial assets and compares the production profiles of those companies with technology and emissions pathways from climate transition scenarios at the sector and/or technology level.
--- a/vignettes/cookbook_overview.Rmd
+++ b/vignettes/cookbook_overview.Rmd
@@ -1,8 +1,8 @@
 ---
-title: "Cookbook - (1) Overview"
+title: "Overview"
 output: rmarkdown::html_vignette
 vignette: >
-  %\VignetteIndexEntry{Cookbook - (1) Overview}
+  %\VignetteIndexEntry{Overview}
   %\VignetteEngine{knitr::rmarkdown}
   %\VignetteEncoding{UTF-8}
 ---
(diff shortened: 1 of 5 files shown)


cjyetman:
I wonder if they even need the "Cookbook - " prefix? Somewhat clear from the section header?

jdhoffa:
> I wonder if they even need the "Cookbook - " prefix? Somewhat clear from the section header?

See the code, I had removed that too haha

jacobvjk:
If we do this, I think it would be good to add an overview at least to the first chapter that describes what is to come, just for user orientation. While I agree that it is implicit from the dropdown in the nav bar, that is not exactly a typical way to present the contents one can expect from a document.. I think the target audience is not necessarily one that hangs out on GH a lot, so I think being a bit more explicit somewhere in the cook book does not hurt

Also, the links at the bottom of the chapters should be renamed as well then

jdhoffa:
That makes sense. Then I will adjust.

jacobvjk:
bit unsure why the macos check keeps failing, the same failure has recently been popping up elsewhere, but re-triggering it has usually solved it.. guess we may just have to wait and see, since it seems to be related to the latest R version

cjyetman:
> bit unsure why the macos check keeps failing, the same failure has recently been popping up elsewhere, but re-triggering it has usually solved it.. guess we may just have to wait and see, since it seems to be related to the latest R version

definitely something widespread going on:
https://github.com/r-lib/actions/issues/950
https://github.com/r-lib/actions/issues/948

Looks like either an intermittent GH Actions network problem, or something to do with the latest GH Actions macOS runners being arm64 and not x86 and possibly not building R or its dependencies properly.

cjyetman:
likely something with the `macos-latest` GH Action runner
https://github.com/r-lib/actions/issues/948#issuecomment-2495473473

supposedly a new version (20241022.361) is currently rolling out, last run here failed with the old version (20241119.509) at 2024.11.23 17:42 GMT+1
https://github.com/RMI-PACTA/pacta.multi.loanbook/actions/runs/11973244713/job/33424388196#step:1:9

more detail likely related: https://github.com/actions/runner-images/issues/10864#issuecomment-2478461845

cjyetman:
macOS-latest failure resolved by GitHub's latest macOS 14 runner image 20241125.556
https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20241125.556

jacobvjk:
> That makes sense. Then I will adjust.

I took the liberty to address this, so that we can close this PR soon @jdhoffa

jdhoffa:
Thank you!! I will leave the rest of this PR to you then.
Feel free to open it from Draft whenever you're ready

vignettes/cookbook_overview.Rmd
```diff
@@ -49,6 +49,14 @@ plot_table <- function(table) {
 
 This cookbook provides a step-by-step guide to running the PACTA for Supervisors analysis using the `pacta.multi.loanbook` package. The analysis is designed to help financial supervisors assess the alignment of banks' loan books with the Paris Agreement goals.
 
+The sections of the cookbook are as follows:
+
+1. **Overview**: This section provides an overview of the PACTA for Supervisors analysis and its main steps.
+2. **Preparatory Steps**: Prepare the ABCD data and optionally a custom sector split.
+3. **Running the Analysis**: Match the raw loan books to the ABCD data and run the PACTA for Supervisors analysis.
+4. **Interpretation of Results**: Interpret the results of the analysis and understand the outputs.
+5. **Advanced Topics**: Learn about advanced topics such as custom grouping of results and how to adjust the analysis parameters for best results on advanced research questions.
```
cjyetman: should these link to the other pages?

cjyetman (approved):
looks good, just one nit comment about making the section descriptions linkable

vignettes/cookbook_overview.Rmd
jacobvjk: might as well, I suppose

cjyetman (approved):

jacobvjk (approved):

PR #279, opened 2024-12-02
use new `sector_classification` argument in `r2dii::match_name()`

### LINKED ISSUE #270: add description to Manual Sector Classification section of cookbook_preparatory_steps
former todo removed in #269 
> **TODO:** Add once we have a stable implementation of this feature

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/855c156e0422e1d5a064804a6ec7d529e954f9b9/vignettes/cookbook_preparatory_steps.Rmd#L151-L158

### LINKED ISSUE #277: integrate new `sector_classification` argument of `r2dii.match::match_name()`
- depends on https://github.com/RMI-PACTA/r2dii.match/pull/509

A new `sector_classification` argument is being added to `r2dii.match::match_name()` to replace the mechanism of using `options(r2dii.match.sector_classifications = own)`. Once this lands, we will need to adapt here, specifically...

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/76e8cbd718fcf7687acb3774dfdf6389ea031d3f/R/match_loanbooks.R#L143-L161

### LINKED ISSUE #286: Bump required version of r2dii.match to >= 0.3.0
relates to https://github.com/RMI-PACTA/pacta.multi.loanbook/pull/279
depends on https://github.com/RMI-PACTA/r2dii.match/issues/510

cjyetman:
- depends on https://github.com/RMI-PACTA/r2dii.match/pull/509
- depends on https://github.com/RMI-PACTA/r2dii.match/issues/510
- depends on r2dii.match v0.3.0 being released/available on CRAN

- closes #277
- closes #286 
- closes #270 

Implement new mechanism for using custom sector classification in `r2dii::match_name()` with the `sector_classification` argument.


jdhoffa:
Woohoo!

R/match_loanbooks.R
```diff
@@ -143,21 +143,16 @@ match_loanbooks <- function(config) {
...
-            abcd = abcd,
-            by_sector = matching_by_sector,
-            min_score = matching_min_score,
-            method = matching_method,
-            p = matching_p,
-            overwrite = matching_overwrite,
-            join_id = matching_join_id
-          )
-        }
+      matched_lbk_i <- r2dii.match::match_name(
```
jacobvjk: nit: instead of this big conditional block here, we could now easily just always pass the argument `sector_classification = sector_classification_system`, if we introduce else logic further above where we load the sector classification system. if no manual file is provided we assign the r2dii.data version.

jacobvjk (approved):
one nit, which I think would make the code more legible, but this all works as expected

R/match_loanbooks.R
cjyetman: totally agree.... I went for easiest to review/minimal change, but also saw that the conditional is not really needed anymore.... I can update this PR, or you can if you want to / get to it

jacobvjk: done. I might update the cookbook while at it

jacobvjk: hm, did not get the error when running `devtools::check()` locally.. Maybe this is not using the latest dev-version of `r2dii.match`?

cjyetman: oh, interesting.... yeah, we're now in that loop of relying on a release version of another package

jacobvjk: yes, unfortunately. which means we need to bump the required version of `r2dii.match` in the `DESCRIPTION` soon too

cjyetman: could temporarily set `Remotes` for it to get the dev version

but one thing I definitely forgot to do here is pin the `r2dii.match` version to >=0.3.0 (?) in DESCRIPTION

cjyetman: realistically, I think the most appropriate approach would be to wait until `r2dii.match` v0.3.0 is released on CRAN, then make sure this PR includes the >=v0.3.0  requirement in DESCRIPTION

jdhoffa: Agree! Let's wait until it's on CRAN to merge this PR

jdhoffa: Relates to https://github.com/RMI-PACTA/r2dii.match/issues/510

jdhoffa:
Converting to draft until https://github.com/RMI-PACTA/r2dii.match/issues/510 is closed so we don't accidentally merge this :-)

cjyetman:
r2dii.match 0.3.0 is released (https://github.com/RMI-PACTA/r2dii.match/issues/510) and checks are now passing, so opening for review

jacobvjk (approved):
still works as expected with the update of r2dii.match
lgtm

PR #291, opened 2024-12-06
grammer fix

cjyetman:


cjyetman:
this is ambiguous, but I now prefer the original... closing

PR #311, opened 2024-12-09
add examples and return value explanations to exported functions

### LINKED ISSUE #272: add examples to exported functions
inline comments removed in #273 and replaced with this issue

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/ccfd776154e462e271d1ad89322209faa8e96bff/R/analyse.R#L17-L21

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/ccfd776154e462e271d1ad89322209faa8e96bff/R/prioritise_and_diagnose.R#L19-L21

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/ccfd776154e462e271d1ad89322209faa8e96bff/R/prepare_abcd.R#L18-L20

https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/ccfd776154e462e271d1ad89322209faa8e96bff/R/match_loanbooks.R#L19-L21

### LINKED ISSUE #309: docs: add explanation of what the side-effects are in the @return section of the function
Rather than `NULL` as we currently have it

cjyetman:
- closes #272
- closes #309

also added description, example, and return value explanation for `initialise_default_project()`

### DIFF
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -3,11 +3,10 @@
 #' @description
 #' `analyse()` runs the necessary steps to analyse the matched loan books,
 #' producing both the outputs of the standard PACTA analysis and the outputs of
-#' the net aggregated alignment metric, including tables and plots.
-#' Parameters for all steps are read from a `config.yml` file. The function is
-#' called for its side effects and writes the prepared and diagnosed data sets
-#' to the directory `output/analysis`, where `output` is the
-#' output directory  specified in the `config.yml`.
+#' the net aggregated alignment metric, including tables and plots. Parameters
+#' for all steps are read from a `config.yml` file. The function is called for
+#' its side effects and writes the prepared and diagnosed data sets in the
+#' directory specified by `dir_analysis` in the `config.yml`.
 #'
 #' `analyse()` and `analyze()` are synonyms.
 #'
@@ -15,9 +14,9 @@
 #'
 #' @return
 #'
-#' `analyse()` returns `NULL` invisibly. It is called for its side-effect of
-#' creating the outputs of the standard PACTA analysis and the outputs of the
-#' net aggregated alignment metric in your `output/analysis` directory.
+#' `analyse()` returns `NULL` invisibly. The function is called for its side
+#' effects and writes the prepared and diagnosed data sets in the directory
+#' specified by `dir_analysis` in the `config.yml`.
 #'
 #' @examples
 #' \dontrun{
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -15,7 +15,7 @@
 #'
 #' @return
 #'
-#' `analyse()` returns NULL invisibly. It is called for its side-effect of
+#' `analyse()` returns `NULL` invisibly. It is called for its side-effect of
 #' creating the outputs of the standard PACTA analysis and the outputs of the
 #' net aggregated alignment metric in your `output/analysis` directory.
 #'
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -13,7 +13,17 @@
 #'
 #' @param config either a path to a config.yml file or a list of parameters
 #'
-#' @return NULL
+#' @return
+#'
+#' `analyse()` returns NULL invisibly. It is called for its side-effect of
+#' creating the outputs of the standard PACTA analysis and the outputs of the
+#' net aggregated alignment metric in your `output/analysis` directory.
+#'
+#' @examples
+#' \dontrun{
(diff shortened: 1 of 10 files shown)


cjyetman:
@jdhoffa is it supposed to look like this on the pkgdown site?
Screenshot 2024-12-09 at 11 59 27

jdhoffa:
Screenshot 2024-12-09 at 12 05 10
Hmm indeed, it seems like this is how it renders in other `pkgdown` site as well. 

That's a bit bizarre

jdhoffa (approved):
looks good!

R/analyse.R
```diff
@@ -13,7 +13,17 @@
 #'
 #' @param config either a path to a config.yml file or a list of parameters
 #'
-#' @return NULL
+#' @return
+#'
+#' `analyse()` returns `NULL` invisibly. It is called for its side-effect of
+#' creating the outputs of the standard PACTA analysis and the outputs of the
+#' net aggregated alignment metric in your `output/analysis` directory.
```
jacobvjk: nit: to be super pedantic, the output directory is whatever the user specifies in `dir_analysis` in the `config.yml`..

R/match_loanbooks.R
```diff
@@ -14,7 +14,17 @@
 #'
 #' @param config either a path to a config.yml file or a list of parameters
 #'
-#' @return NULL
+#' @return
+#'
+#' `match_loanbooks()` returns `NULL` invisibly. It is called for its
+#' side-effect of creating the matched loan books in your `output/match`
```
jacobvjk: nit: see above

R/prepare_abcd.R
```diff
@@ -12,7 +12,16 @@
 #'
 #' @param config either a path to a config.yml file or a list of parameters
 #'
-#' @return NULL
+#' @return
+#'
+#' `prepare_abcd()` returns `NULL` invisibly. It is called for its side-effect
+#' of creating the prepared ABCD file in your `output/prepare` directory.
```
jacobvjk: nit: see above

R/prioritise_and_diagnose.R
```diff
@@ -13,7 +13,17 @@
 #'
 #' @param config either a path to a config.yml file or a list of parameters
 #'
-#' @return NULL
+#' @return
+#'
+#' `prioritise_and_diagnose()` returns `NULL` invisibly. It is called for its
+#' side-effect of creating the prepared and diagnosed data sets in your
+#' `output/prioritise_and_diagnose` directory.
```
jacobvjk: nit: see above

jacobvjk (approved):
a few nits, be can also be addressed later.

generally I find this format a bit confusing with the `if (FALSE)` but it seems like that is not fully our choice?
Screenshot 2024-12-09 at 12 15 46

jdhoffa:
> Screenshot 2024-12-09 at 12 05 10 Hmm indeed, it seems like this is how it renders in other `pkgdown` site as well.

Yeah so this seems to be an artifact of:
https://github.com/r-lib/pkgdown/commit/a133692bde19ed2f750b85b8c0021f4cf068eac8
(related issue: https://github.com/r-lib/pkgdown/issues/29)

Basically, they implemented functionality to force the rendering of `dontrun` sections in `pkgdown` (for some reason?)
By default, this is set to `FALSE`, and I guess under the hood it probably is a macro that injects this `if(FALSE) {...}` control flow chunk around the `dontrun` code.

Nothing we can do about this on our end.

R/analyse.R
cjyetman: hmmmm, initially I was thinking since this is text the CRAN folks will likely review closely, let's not make it too complicated, but maybe it's not too complicated and more correct to say...
> `analyse()` returns `NULL` invisibly. It is called for its side-effect of creating the outputs of the standard PACTA analysis and the outputs of the net aggregated alignment metric in the directory specified by `dir_analysis` in your config file.

does that sound good/right @jacobvjk?

jacobvjk: that sounds more correct, but admittedly I am not sure what the people at CRAN would prefer. I think we just try one and worst case we will have to adjust it

cjyetman: ok, I'll go with "more correct" for now... changes incoming

cjyetman:
round 2... I've specified the config params that determine the output directories and I've aligned the text in the descriptions with the text in the return value descriptions

jacobvjk (approved):

PR #313, opened 2024-12-09
update default config for 2023

### LINKED ISSUE #299: consider using 2023 scenarios in the default example
https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/28c1b676782653b156a0d14781625a65ba8a027c/inst/default_config.yml#L9-L20

- related #220

cjyetman:
- closes #299

### DIFF
--- a/inst/default_config.yml
+++ b/inst/default_config.yml
@@ -6,17 +6,17 @@ default:
     dir_prioritized_loanbooks_and_diagnostics: "path/to/prioritized_loanbooks_and_diagnostics"
     dir_analysis: "path/to/analysis"
   file_names:
-    filename_scenario_tms: "scenarios_2022_tms.csv"
-    filename_scenario_sda: "scenarios_2022_sda.csv"
-    filename_abcd: "abcd.xlsx"
+    filename_scenario_tms: "2023 IEA JRC ISF Auto Fossil Fuel Power (start year 2023).csv"
+    filename_scenario_sda: "2023 CO2 Intensity IEA JRC ISF Steel Cement Aviation (start year 2023).csv"
+    filename_abcd: "2023-02-17_AI_RMI_PACTA for Banks Free dataset_EO_2022Q4.xlsx"
     sheet_abcd: "Company Indicators - PACTA Comp"
   project_parameters:
-    scenario_source: "weo_2022"
+    scenario_source: "weo_2023"
     scenario_select: "nze_2050"
     region_select: "global"
     # normally the start year should correspond with year of the publication of
     # the scenario in use
-    start_year: 2022
+    start_year: 2023
     time_frame: 5
     by_group: "group_id"
   sector_split:


inst/default_config.yml
```diff
@@ -6,17 +6,17 @@ default:
...
     sheet_abcd: "Company Indicators - PACTA Comp"
   project_parameters:
-    scenario_source: "weo_2022"
+    scenario_source: "weo_2023"
     scenario_select: "nze_2050"
     region_select: "global"
     # normally the start year should correspond with year of the publication of
     # the scenario in use
-    start_year: 2022
+    start_year: 2023
```
jdhoffa, line 19: Is this the only place that it needs to change? 
I guess so, we decided to keep all the cookbook references as is right, in order to not regenerate the sample plots?

cjyetman: yeah, I think it's ok if the docs don't use the same values. The values in the docs/cookbook are more informative and in line with the plots, but values here are what is created for the user when using `initialise_default_project()`.

@jacobvjk?

jacobvjk: we could update it in `config_yml.Rmd` as well. The cookbook itself does not mention the publication year directly

jacobvjk: overall no strong feelings about this, I would not adjust the cookbook itself and I think the config vignette is a secondary consideration, because it is meant to be more general information

cjyetman: I think that 
``` yaml
filename_scenario_sda: "scenarios_2022_ei_p4b.csv"
```
is more informative for the documentation than
``` yaml
filename_scenario_sda: "2023 CO2 Intensity IEA JRC ISF Steel Cement Aviation (start year 2023).csv"
```
but I could change it to 
``` yaml
filename_scenario_sda: "scenarios_2023_ei_p4b.csv"
```
although that makes it out of sync with the plots in the cookbook... is "scenarios_2023_ei_p4b.csv" preferred?

jacobvjk: let's keep it as it is, you are right. Being overspecific here looks weird.

jacobvjk (approved):

jdhoffa:
@jacobvjk @cjyetman FYI this seems to have broken the testing worfklow for Daisy and Matze (as they are using the dummy data provided by @jacobvjk that is all set to `2022`)...

jdhoffa:
I would propose that we revert this PR, and re-open the issue (and tackle it _after_ all our internal folks have completed their testing)

jdhoffa:
Otherwise, we will need to generate and send them a new batch of testing datasets

cjyetman:
> I would propose that we revert this PR, and re-open the issue (and tackle it _after_ all our internal folks have completed their testing)

and ask them to re-install (versus adjust the config parameters as the cookbook explains to do)? not arguing, just seems like the damage is already done and reverting this just causes more damage rather than erases previous damage

jdhoffa:
True, then the alternative would be to just provide `2023` scenarios? 

(also, the error message that they get when the scenarios don't match the start_year is confusing, and not super informative. We may want to improve exception handling there)

cjyetman:
2023 scenario files are here, no? that's what I used anyway... and presumably that's what a completely new user would do that doesn't benefit from having us prepare a zip file for them?

> Since RMI has taken over stewardship of PACTA, the prepared scenario files can also be accessed as CSV downloads on the [PACTA website](https://pacta.rmi.org/pacta-for-banks-2020/) under the “Methodology and Documents” tab of the “PACTA for Banks” section. The files are usually updated annually based on the latest scenario publications and as a general rule, the year of the publication defines the initial year of the scenario data set. This is commonly also used as the start year of the analysis, which is specified in the config.yml file.

cjyetman:
or they could use the 2022 scenarios that are in the zip and set the config parameters appropriately? honestly, that seems like the ideal... that's what the review process is for, to test if a user can do it according to the instructions?

cjyetman:
ah crap.... but the ABCD data has to be aligned 🤔

PR #321, opened 2024-12-09
Optimize plots in cookbook

### LINKED ISSUE #253: feat: optimize plots and visuals for the cookbook
User Story: 
* As a reader of the cookbook, I would like to easily see and understand all visuals

Acceptance Criteria:
- [ ] All titles, labels, etc. in all plots should be larger than or equal to the size of the text content in the cookbook (e.g. in screenshot below, all plot font sizes should be increased to at least the size of "An example plot..." 

Screenshot 2024-11-20 at 12 36 48

jacobvjk:
closes #253 

I am sure one could spend more time on this to make the plots even more readable, but I think with an adequate amount of time, this is what I can produce


jdhoffa (approved):
Much better!

cjyetman (approved):

PR #337, opened 2024-12-11
Validate matching timestamps

jacobvjk:
closes #327 

- gains validation that entire period between parameters `start_year` and `start_year + time_frame` is available in abcd and, where necessary in AI advanced company indicators
- gains validation that parameter `start_year` lines up with the initial years of the tms and sda scenario inputs

PR #338, opened 2024-12-11
use demo data in test-4-analyze

jacobvjk:
this is an extension of https://github.com/RMI-PACTA/pacta.multi.loanbook/pull/337

I attempt to actually use the demo data in test-4-analyze by changing the test config accordingly.
The code checks if the config uses the demo_2020 scenario source and if so, it will load the demo data from for regional matching from r2dii.data.

This is still not functional, because there is no common scenario name available in r2dii.data::r2dii.data::scenario_demo_2020 and in r2dii.data::r2dii.data::co2_intensity_scenario_demo, so there is no option of passing the "correct" scenario name in the test config..

fyi @cjyetman  @jdhoffa

### DIFF
--- a/R/run_aggregate_alignment_metric.R
+++ b/R/run_aggregate_alignment_metric.R
@@ -51,11 +51,20 @@ run_aggregate_alignment_metric <- function(config) {
   assert_inherits(time_frame, "integer")
 
   # load input data----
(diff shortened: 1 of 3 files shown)


jacobvjk:
replaced by #339

Issue #349, opened 2024-12-12
Release pacta.multi.loanbook 0.1.0


jacobvjk:
First release

- [ ] [usethis::use_news_md()](https://usethis.r-lib.org/reference/use_news_md.html) _(not required for initial GH release)_
- [ ] [usethis::use_cran_comments()](https://usethis.r-lib.org/reference/use_cran_comments.html) _(not required for initial GH release)_
- [x] Update (aspirational) install instructions in README
- [x] Proofread Title: and Description:
- [x] Check that all exported functions have @returns and @examples
- [x] Check that Authors@R: includes a copyright holder (role ‘cph’)
- [x] Check licensing of included files
- [x] Review https://github.com/DavisVaughan/extrachecks

Prepare for release:

- [x] git pull
- [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_r2dii.match.html)
- [x] Check if any deprecation processes should be advanced, as described in [Gradual deprecation](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation)
- [ ] [Polish NEWS](https://style.tidyverse.org/news.html#news-release) _(not required for initial GH release)_
- [x] urlchecker::url_check()
- [x] devtools::build_readme()
- [x] devtools::check(remote = TRUE, manual = TRUE)
- [x] devtools::check_win_devel()
- [x] revdepcheck::revdep_check(num_workers = 4)
- [ ] Update cran-comments.md _(not required for initial GH release)_
- [x] git push
- [x] Draft blog post

Submit to CRAN:

- [x] usethis::use_version('minor')
- [ ] devtools::submit_cran() _(not required for initial GH release)_
- [ ] Approve email _(not required for initial GH release)_

Wait for CRAN...

- [ ] Accepted 🎉 _(not required for initial GH release)_
- [x] git push
- [x] Finish & publish blog post
- [x] Add link to blog post in pkgdown news menu
- [x] usethis::use_github_release()
- [x] usethis::use_dev_version(push = TRUE)
- [x] Share on social media

jacobvjk:
@jdhoffa It seems like it is not too common to have anything particular in the NEWS.md for the first CRAN submission. I checked some tidyverse examples and their NEWS tend to be populated starting with releases > 0.1.0.

jdhoffa:
Ah indeed! NEWS should be empty. 

The changelog starts after the package is on CRAN

[...]

PR #352, opened 2024-12-12
Prepare Release 0.1.0

jacobvjk:
relates to #349


jacobvjk:
the remaining NOTE that triggers the check failures here does not seem to be an issue, as it will be resolved by bumping the DESCRIPTION version, see https://stackoverflow.com/questions/36701433/r-package-building-with-devtoolsbuild-win-version-contains-large-components

jdhoffa (approved):

PR #355, opened 2024-12-12
set `_R_CHECK_CRAN_INCOMING_` to `TRUE` in CI for more robust testing

cjyetman:
should see similar errors discovered with `devtools::check_win_devel()`

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -64,3 +64,5 @@ jobs:
           _R_CHECK_CRAN_INCOMING_: true
           _R_CHECK_CRAN_INCOMING_REMOTE_: true
           _R_CHECK_CRAN_INCOMING_USE_ASPELL_: true
+          _R_CHECK_URLS_SHOW_301_STATUS_: true
+          _R_CHECK_RD_LINE_WIDTHS_: true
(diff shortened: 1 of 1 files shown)


jacobvjk (approved):
seems good in principle, but it does not seem to find the bug you addressed in #353 
the other NOTEs are already addressed in #352 so I would expect to see this pass once this is merged and the release branch is updated

cjyetman:
Yeah, I don't necessarily expect to have this merged. I would suggest merging everything else you're gonna merge, then come back here and update this branch and check the test results before running another check_win_devel. 

I don't know why this doesn't trigger the test failure on Windows path / invalid regex. That's what I was going for. Might be additional env vars to set to faithfully replicate what check_win_devel does. If we can sync this up with that, then I would consider merging this.

cjyetman:
@jacobvjk or you could (temporarily?) add this hack to your #352 branch to test it before merging/check_win_devel

cjyetman:
I propose closing this PR and eventually considering merging #365, which would put this CRAN incoming check in its own GH Action, enabling it to be easier to determine when/how it gets triggered, and its much more feature-rich in how it can report issues to us.

cjyetman:
superseded by #365

PR #357, opened 2024-12-12
use canonical URL of the CRAN page for a package

cjyetman:
otherwise CRAN might complain


jdhoffa (approved):

jacobvjk:
this is already handled in #352

jacobvjk:
Also we cannot link to the canonical URL either at this point, since it does not exist before the package is on CRAN, so url_checker will still complain

cjyetman:
Oops... closing

PR #365, opened 2024-12-16
Create R-CMD-check-CRAN-incoming.yaml

cjyetman:
Adds a new R-CMD-check-CRAN-incoming.yaml GH Action

The Action attempts to run `R CMD check` in the same way that [Win Builder](https://win-builder.r-project.org) does, enabling one to get early notification of problems that will ultimately need to be addressed before a CRAN submission. The goal is to replicate the process used on CRAN/Win Builder (which is probably not entirely possible) and make appropriate notifications when something of interest happens.

To that end, it...
- runs `R CMD check` with arguments, environment variables, etc. as close as possible to those run on CRAN and Win Builder
- parses the results for the notes directly following `* checking CRAN incoming feasibility ... [12s] NOTE`
- parses those notes to remove notes that will always be there and/or do not suggest a needed action, e.g. `Maintainer: 'CJ Yetman <cj@cjyetman.com>'` and `New submission`
- if there are notes left after the parsing
  - throws an error on the GH Action
  - prints the notes causing the error in the GH Action log
  - prints the entire `00check.log` as a [Job Summary](https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/)
  - uploads the `00check.log` as an [Artifact](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow)
  - adds a comment (or updates the previous comment) in the PR thread showing the CRAN incoming feasibility notes that were discovered

I'm still questioning whether it should also parse out the `Version contains large components (0.0.0.9000)` note. That would effectively show up on every PR unless the version number was a proper release version number, but when a release is being made it would be a good nudge that the version number needs to be changed. Thoughts on that?

It could alternatively only be run manually through `workflow_dispatch` and/or it could ignore the `Version contains large components (0.0.0.9000)` note.


jacobvjk:
I tend to think that large components in the version number should not throw an error here, because that will be the standard situation during most of the development phase. Since bumping the version number is part of the CRAN release checklist, I believe we will not miss that when submitting to CRAN in all likelihood

jacobvjk (approved):

PR #366, opened 2024-12-17
use a mermaid flowchart

cjyetman:
builds a flowchart programmatically, which is easier to adapt and avoids adding a large static image

does add a dependency on `DiagrammeR`, but it's a Suggest not an Import

---

render as (further customization could be done, e.g. the function names could be made links and I think color/shape/size can be controlled)

Screenshot 2024-12-19 at 10 55 16

https://rmi-pacta.github.io/pacta.multi.loanbook/pr-preview/pr-366/articles/cookbook_running_the_analysis.html#structure-of-the-workflow

versus

Screenshot 2024-12-19 at 10 57 11

https://rmi-pacta.github.io/pacta.multi.loanbook/articles/cookbook_running_the_analysis.html#structure-of-the-workflow


jacobvjk (changes requested):
Getting rid of a large png makes sense to me.
I would suggest we adjust the color to RMI color scheme and make this a bit larger though, I currently find this a bit delicate in size

jacobvjk (approved):
lgtm

Issue #368, opened 2024-12-18
Release pacta.multi.loanbook 0.1.1 on CRAN


jacobvjk:
First release

- [x] [usethis::use_news_md()](https://usethis.r-lib.org/reference/use_news_md.html)
- [x] [usethis::use_cran_comments()](https://usethis.r-lib.org/reference/use_cran_comments.html)
- [x] Update (aspirational) install instructions in README
- [x] Proofread Title: and Description:
- [x] Check that all exported functions have @returns and https://github.com/examples
- [x] Check that Authors@R: includes a copyright holder (role ‘cph’)
- [x] Check licensing of included files
- [x] Review https://github.com/DavisVaughan/extrachecks

Prepare for release:

- [x] git pull
- [x] Check [current CRAN check results](https://cran.rstudio.org/web/checks/check_results_r2dii.match.html)
- [x] Check if any deprecation processes should be advanced, as described in [Gradual deprecation](https://lifecycle.r-lib.org/articles/communicate.html#gradual-deprecation)
- [x] [Polish NEWS](https://style.tidyverse.org/news.html#news-release)
- [x] urlchecker::url_check()
- [x] devtools::build_readme()
- [x] devtools::check(remote = TRUE, manual = TRUE)
- [x] devtools::check_win_devel()
- [x] revdepcheck::revdep_check(num_workers = 4)
- [x] Update cran-comments.md
- [x] git push
- [x] Draft blog post

Submit to CRAN:

- [x] usethis::use_version('minor')
- [x] devtools::submit_cran()
- [x] Approve email

Wait for CRAN...

- [x] Accepted 🎉
- [x] git push
- [x] Finish & publish blog post
- [x] Add link to blog post in pkgdown news menu
- [x] usethis::use_github_release()
- [x] usethis::use_dev_version(push = TRUE)
- [x] Share on social media

succeeds #349

cjyetman:
- related #210 
- related #148

PR #379, opened 2025-01-08
Release version 0.1.1

jacobvjk:
relates to #368

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -10,3 +10,4 @@
 ^codecov\.yml$
 ^cran-comments\.md$
 ^README\.Rmd$
+^CRAN-SUBMISSION$
--- a/CRAN-SUBMISSION
+++ b/CRAN-SUBMISSION
@@ -1,0 +1,3 @@
+Version: 0.1.1
+Date: 2025-01-08 13:47:18 UTC
+SHA: 5d3bf7171a53641d9622d85a2b637fa6ecc5f7f0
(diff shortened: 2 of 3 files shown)


jacobvjk:
@cjyetman  I believe this seems to be causing trouble here? https://github.com/RMI-PACTA/pacta.multi.loanbook/commit/9d940a37ef1085302151670b1913e31504a34d13

from what I understand [we do use JamesIves to deploy our github pages](https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/2b31618cd38353a6eb260fc1f7c5fef1ecd43ea9/.github/workflows/pkgdown.yaml#L44), yet [this suggested solution](https://github.com/rossjrw/pr-preview-action?tab=readme-ov-file#ensure-your-main-deployment-is-compatible) does not seem to apply, because we [disabled the cleanup](https://github.com/RMI-PACTA/pacta.multi.loanbook/blob/2b31618cd38353a6eb260fc1f7c5fef1ecd43ea9/.github/workflows/pkgdown.yaml#L48) in any case?

If my reading is correct, I am a bit unsure where to look for fixing this, without spending a lot of time on it. Maybe you have a hunch?

jacobvjk:
I think this was accidentally closed via #380. Will reopen

jdhoffa (approved):
Given that the failing CRAN check pertains only to the version number, I think this is good to go!

PR #380, opened 2025-01-11
deploy preview sites under `docs` directory

cjyetman:
- related #379 
- fixes https://github.com/RMI-PACTA/pacta.multi.loanbook/pull/379#issuecomment-2582176196


jacobvjk (approved):
given that the failing check is the CRAN incoming check that seems to complain about this version not matching the version on CRAN, I would assume the problem should be resolved once we merge #379 

so this lgtm

